PR #29, opened 2023-06-08
feat(APIM-347): create an authenticated microsoft graph client

AlexBramhill:
**Introduction**
We require Microsoft Graph authentication to create many of the ESTORE endpoints. We need to create a method to authenticate with Microsoft Graph prior to undertaking these tickets

**Resolution**
We authenticate with Microsoft Graph via [MSAL](https://www.npmjs.com/package/@azure/msal-node).

**Misc**
I've kept this ticket reduced in scope -- no caching or logging. I am assuming this will be covered in later tickets.
The testing is a bit interesting -- I feel there might be a better way?
I've left a few comments in the code pointing out any oddities
I've renamed the envs


src/config/index.ts
```diff
@@ -1,4 +1,5 @@
 import AppConfig from './app.config';
 import DocConfig from './doc.config';
+import graphAuthenticationConfig from './graph-authentication.config';
```
lmarrai-sw: Can we use UpperCamelCase please?

test/support/environment-variables.ts
```diff
@@ -9,6 +9,14 @@ export const ENVIRONMENT_VARIABLES = Object.freeze({
   SWAGGER_USER: valueGenerator.string(),
   SWAGGER_PASSWORD: valueGenerator.string(),
   API_KEY: valueGenerator.string(),
+
+  // These are not currently used in graph-authentication.service.test.ts
```
lmarrai-sw: (If we're using this the same as we did on TFS then) this file is for the environment variables for the API tests, not for unit tests

src/modules/sites/sites.service.ts
```diff
@@ -2,14 +2,14 @@ import { HttpService } from '@nestjs/axios';
 import { Injectable } from '@nestjs/common';
 import { lastValueFrom } from 'rxjs';
 
-import { GraphAuthService } from '../auth/graph-auth.service';
+import { GraphAuthenticationService } from '../graph-authentication/graph-authentication.service';
```
lmarrai-sw: Can we use `@ukef/modules` import please?

src/modules/sites/sites.module.ts
```diff
@@ -1,12 +1,14 @@
...
 import { Module } from '@nestjs/common';
 
-import { GraphAuthService } from '../auth/graph-auth.service';
+import { GraphAuthenticationModule } from '../graph-authentication/graph-authentication.module';
 import { SharePointController } from './sites.controller';
 import { SiteExistsService } from './sites.service';
+
 @Module({
-  imports: [HttpModule],
+  // Will look to move GraphAuthenticationModule to a shared Graph module
```
lmarrai-sw: Might be worth sticking `TODO` at the start of the comment?

src/modules/graph-authentication/graph-authentication.service.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  getAccessToken = async (): Promise<string> => {
+    const { accessToken } = await this.cca
+      .acquireTokenByClientCredential({
+        scopes: [this.config.scope],
+      })
+      .catch(() => {
+        throw new GraphAuthenticationFailedException('Failed to get graph token');
+      });
```
lmarrai-sw: Do we want to pass the caught error in as the `innerError`?

lmarrai-sw: (and test for that in the unit tests)

src/modules/graph-authentication/graph-authentication.service.test.ts
```diff
@@ -0,0 +1,57 @@
...
+    authority,
+    scope,
+  };
+
+  let graphAuthenticationService: GraphAuthenticationService;
+  let acquireTokenByClientCredentialSpy: jest.SpyInstance<Promise<AuthenticationResult>, [request: ClientCredentialRequest], any>;
+
+  beforeEach(() => {
+    graphAuthenticationService = new GraphAuthenticationService(graphAuthenticationConfig);
+    acquireTokenByClientCredentialSpy = jest.spyOn(graphAuthenticationService['cca'], 'acquireTokenByClientCredential');
```
lmarrai-sw: I don't think we should be spying on a `private` field of the class, it makes the test too implementation-dependant. Let me know if you want to have a chat about a different approach

src/modules/graph-authentication/graph-authentication.service.ts
```diff
@@ -0,0 +1,36 @@
...
+    return accessToken;
+  };
+}
```
lmarrai-sw: Did we consider using `@azure/identity` at all? It looks like it works really well with `@microsoft/microsoft-graph-client` and we could skip getting the access token entirely (like [so](https://learn.microsoft.com/en-gb/graph/sdks/choose-authentication-providers?tabs=Javascript#client-credentials-provider))- or do we need more control over the authentication?

src/modules/graph/graph.service.ts
```diff
@@ -0,0 +1,27 @@
...
+type RequiredConfigKeys = 'clientId' | 'clientSecret' | 'tenantId' | 'scope';
+
+@Injectable()
+export class GraphService {
+  client: Client;
+  constructor(
+    @Inject(GraphConfig.KEY)
+    private readonly config: Pick<ConfigType<typeof GraphConfig>, RequiredConfigKeys>,
+  ) {
+    const credential = new ClientSecretCredential(this.config.tenantId, this.config.clientId, this.config.clientSecret);
```
lmarrai-sw: I think if we had the prefer-destructuring lint rule turned on, it would ask you to destructure these values from the config instead

src/modules/graph/graph.service.ts
```diff
@@ -0,0 +1,27 @@
...
+import GraphConfig from '@ukef/config/graph.config';
+
+type RequiredConfigKeys = 'clientId' | 'clientSecret' | 'tenantId' | 'scope';
+
+@Injectable()
+export class GraphService {
+  client: Client;
+  constructor(
+    @Inject(GraphConfig.KEY)
+    private readonly config: Pick<ConfigType<typeof GraphConfig>, RequiredConfigKeys>,
```
lmarrai-sw: I don't think we need `private readonly` here as we don't need to keep the `config` as a field. That also means you could destructure the values immediately if you want

lmarrai-sw (dismissed):
I've suggested 2 minor changes but am happy otherwise :)

src/modules/graph/graph.service.ts
AlexBramhill: Great -- And that means I can also remove the `Pick` as it's unneeded

lmarrai-sw (approved):
LGTM, thanks @AlexBramhill

PR #34, opened 2023-06-14
feat: add facilityId to term store

IgnatG:
TODO: add tests


src/constants/index.ts
```diff
@@ -3,6 +3,12 @@
  * Following constants are served:
  *
  * 1. Auth
+ * 2. EXAMPLES
```
AlexBramhill, line 6: Nice -- I've added EXAMPLES on my branch too (which is now in main) so a rebase might be needed

src/modules/graph/graph.service.ts
```diff
@@ -36,6 +37,11 @@ export class GraphService {
 
     return await request.get();
   }
+
+  post<T>({ path, listItem }: any): Promise<T> {
+    const request = this.client.api(path);
```
AlexBramhill, line 42: Nice - Again, I've added error handling example (although I imagine it will be changed out at a later date). Its on main now

src/modules/terms/terms.service.test.ts
```diff
@@ -0,0 +1,71 @@
...
+      const error = { statusCode: 400 };
+      when(mockGraphServicePost).calledWith(expect.any(Object)).mockRejectedValue(error);
+
+      const result = await service.postFacilityToTermStore(facilityId);
+
+      expect(graphService.post).toHaveBeenCalled();
+      expect(result).toEqual({ message: RESPONSE.FACILITY_TERM_EXISTS });
+    });
+
+    it('should throw an error when the graphService post method throws an error', async () => {
```
AlexBramhill, line 63: I think this might change slightly with the new error handling. I've only scaffolded it for the endpoint I was working on for speed of that ticket's delivery -- However, I think we should eventually move towards a more scalable solution.

src/modules/terms/terms.service.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    try {
+      await this.graphService.post<any>({
+        path: `sites/${ukefSharepointName}:/sites/${tfisSiteName}:/lists/${tfisTermStoreId}/items`,
+        listItem,
+      });
+      return { message: RESPONSE.FACILITY_TERM_CREATED };
+    } catch (error: any) {
+      if (error.statusCode === 400) {
+        return { message: RESPONSE.FACILITY_TERM_EXISTS };
```
AlexBramhill, line 34: Are we sure that 400 is only thrown when the facility term exists? We might want to check the error message (sometimes graph throws the real error in the `code` value) and then either `return { message: RESPONSE.FACILITY_TERM_EXISTS };` or throw a general GraphUnexpectedException

IgnatG: good shout. currently we are returning 200, even when the term already exists in the database. 
I'll add a message on slack to ask the others for thoughts (400 or 200 error)

IgnatG: in the meantime, I've changed it so it's using your error handler. + created a new PR since this was merging against the wrong branch

PR #35, opened 2023-06-15
feat(APIM-134): migrate GET `/site/exist` endpoint

AlexBramhill:
# Introduction
We have a `POST` `/site/exist` endpoint that is to be migrated to `GET` `/site?exporterName={exporterName}`. Note the change to a `GET`, and updated  url.

This endpoint takes an `exporterName` query param and returns either a `200` `202` or `424` depending on the site status (more info below), with the following body: 
```
{
  siteName = string, required [this should be the URL of the list item that is found]
  status = string, required [this should be the Sitestatus of the list item that is found]
}
```

Response codes:
200: site is listed and has been created
202: site is listed but is still waiting to be created
400: bad request
401: unauthorized
404: site does not exist (returns an empty object)
424: site is listed but failed to be created
500: unexpected error

# Resolution

I've reworked the graph service a fair amount -- adding a `get` method to this to allow for general error handling.
I've created a site service and controller to handle this new endpoint
I've utilised the `@Res` decorator to handle the myriad of response codes for this endpoint.

# Misc
Graph tests are to come later
Api-tests are in progress

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -39,6 +39,7 @@
     "nestjs",
     "pino",
     "probabilityof",
+    "tfis",
     "estore",
     "typescript",
     "ukef",
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -1,5 +1,6 @@
 import AppConfig from './app.config';
 import DocConfig from './doc.config';
 import GraphConfig from './graph.config';
+import SharepointConfig from './sharepoint.config';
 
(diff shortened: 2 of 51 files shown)


IgnatG (dismissed):
Happy overall. any tidying up can be done afterwards

avaitonis (approved):
LGTM

PR #38, opened 2023-06-16
feat(APIM-134): migrate get site exists endpoint

AlexBramhill:
# Introduction
We have a `POST` `/site/exist` endpoint that is to be migrated to `GET` `/site?exporterName={exporterName}`. Note the change to a `GET`, and updated  url.

This endpoint takes an `exporterName` query param and returns either a `200` `202` or `424` depending on the site status (more info below), with the following body: 
```
{
  siteName = string, required [this should be the URL of the list item that is found]
  status = string, required [this should be the Sitestatus of the list item that is found]
}
```

Response codes:
200: site is listed and has been created
202: site is listed but is still waiting to be created
400: bad request
401: unauthorized
404: site does not exist (returns an empty object)
424: site is listed but failed to be created
500: unexpected error

# Resolution

This ticket adds the missing testing for the previously merged work on the ticket, including reworks around graph service, abstracting the graph client out into a new service.


src/modules/graph/graph.test-parts/with-common-graph-exception-handling-tests.ts
```diff
@@ -1,6 +1,6 @@
 import { GraphError } from '@microsoft/microsoft-graph-client/lib/src/GraphError';
 
-import { commonGraphExceptionTestCases } from '../common/test-parts/common-graph-exception-handling-test-parts';
+import { commonGraphExceptionTestCases } from '../../../../test/common-test-cases/common-graph-exception-handling-test-cases';
```
IgnatG: is this something that can be converted to an alias-path? i.e. `@ukef/tests/etc`

src/modules/graph/common/common-graph-exception-handling.test.ts
```diff
@@ -1,16 +1,16 @@
 import { GraphError } from '@microsoft/microsoft-graph-client';
 
+import { commonGraphExceptionTestCases } from '../../../../test/common-test-cases/common-graph-exception-handling-test-cases';
```
lmarrai-sw: Can we import from '@ukef/...` instead please?

src/modules/graph/graph.module.ts
```diff
@@ -1,7 +1,7 @@
 import { Module } from '@nestjs/common';
 
 import { GraphClientModule } from '../graph-client/graph-client.module';
-import { GraphService } from './graph.service';
+import GraphService from './graph.service';
```
lmarrai-sw: Why did this change to a default import?

test/docs/get-docs.api-test.ts
```diff
@@ -4,7 +4,7 @@ describe('GET /openapi/yaml', () => {
   let api: Api;
 
   beforeAll(async () => {
-    api = await Api.create();
+    ({ api } = await Api.create());
```
lmarrai-sw, line 7: Do we need the round brackets around the outside?

test/docs/get-docs.api-test.ts
```diff
@@ -26,7 +26,7 @@ describe('GET /openapi/json', () => {
...
```
lmarrai-sw, line 29: Same question as above

test/docs/__snapshots__/get-docs.api-test.ts.snap
```diff
@@ -18,7 +58,21 @@ components:
...
       name: x-api-key
-  schemas: {}
+  schemas:
+    GetSiteStatusByExporterNameResponse:
+      type: object
+      properties:
+        siteId:
+          type: string
+          readOnly: true
+          example: '12345678'
```
lmarrai-sw, line 68: Just to check, does this example have the right number of digits?

src/modules/graph/graph.service.ts
```diff
@@ -9,7 +9,7 @@ export class GraphService {
   private readonly client: Client;
 
   constructor(graphClientService: GraphClientService) {
-    this.client = graphClientService.getClient();
+    this.client = graphClientService.client;
```
lmarrai-sw, line 12: Why did we swap from a getter to a public field?

test/common-tests/common-graph-exception-handling-api-tests.ts
```diff
@@ -0,0 +1,38 @@
...
+  givenRequestWouldOtherwiseSucceed: () => void;
+  givenGraphServiceCallWillThrowError: (error: Error) => void;
+  makeRequest: () => supertest.Test;
+}) => {
+  describe('withCommonGraphExceptionHandling', () => {
+    const graphStatusCode = 0;
+    const graphErrorMessage = 'GraphErrorMessage';
+
+    it.each(commonGraphExceptionTestCases)(
+      'returns a 500 with message "Internal server error" if Microsoft Graph responds with code $graphErrorCode ',
```
lmarrai-sw: There's a blank space at the end of this test name

test/common-tests/common-graph-exception-handling-api-tests.ts
```diff
@@ -0,0 +1,38 @@
...
+export const withCommonGraphExceptionHandlingTests = ({
+  givenRequestWouldOtherwiseSucceed,
+  givenGraphServiceCallWillThrowError,
+  makeRequest,
+}: {
+  givenRequestWouldOtherwiseSucceed: () => void;
+  givenGraphServiceCallWillThrowError: (error: Error) => void;
+  makeRequest: () => supertest.Test;
+}) => {
+  describe('withCommonGraphExceptionHandling', () => {
```
lmarrai-sw: Can we make the name a bit more human-readable please?

test/common-tests/common-graph-exception-handling-api-tests.ts
```diff
@@ -0,0 +1,38 @@
+import { GraphError } from '@microsoft/microsoft-graph-client';
+import supertest from 'supertest';
+
+import { commonGraphExceptionTestCases } from '../common-test-cases/common-graph-exception-handling-test-cases';
+
+export const withCommonGraphExceptionHandlingTests = ({
```
lmarrai-sw: Nice work here!

test/docs/__snapshots__/get-docs.api-test.ts.snap
```diff
@@ -1,10 +1,50 @@
...
+paths:
+  /api/v1/sites:
+    get:
+      operationId: SiteController_getSiteStatusByExporterName
+      summary: Get the site status by exporter name
+      parameters:
+        - name: exporterName
+          required: true
+          in: query
+          description: Name of case site
```
lmarrai-sw: Can we update the description to
```
The name of the exporter used in the deal.
```
please, to match APIM-454?

test/docs/__snapshots__/get-docs.api-test.ts.snap
```diff
@@ -18,7 +58,21 @@ components:
...
+      type: object
+      properties:
+        siteId:
+          type: string
+          readOnly: true
+          example: '12345678'
+        status:
+          type: string
+          readOnly: true
+          example: Created
```
lmarrai-sw, line 72: Should this not have been an enum?

src/modules/site/site.service.test.ts
```diff
@@ -11,9 +12,9 @@ jest.mock('../graph/graph.service');
 describe('SiteService', () => {
   const valueGenerator = new RandomValueGenerator();
 
-  const ukefSharepointName = valueGenerator.string();
-  const tfisSiteName = valueGenerator.string();
-  const tfisListId = valueGenerator.string();
+  const ukefSharepointName = ENVIRONMENT_VARIABLES.SHAREPOINT_MAIN_SITE_NAME;
+  const tfisSiteName = ENVIRONMENT_VARIABLES.SHAREPOINT_TFIS_SITE_NAME;
+  const tfisListId = ENVIRONMENT_VARIABLES.SHAREPOINT_TFIS_LIST_ID;
```
lmarrai-sw: I don't think we want to make this change for the unit tests - the unit tests should be making sure that the service gets these values from the constructor, but since we're using special values now we don't know if the service is using the values from the constructor or if it is importing the constants from a different file

test/common-test-cases/common-graph-exception-handling-test-cases.ts
```diff
@@ -0,0 +1,22 @@
+import { GraphAuthenticationFailedException } from '../../src/modules/graph/exception/graph-authentication-failed.exception';
+import { GraphInvalidRequestException } from '../../src/modules/graph/exception/graph-invalid-request.exception';
+import { GraphUnexpectedException } from '../../src/modules/graph/exception/graph-unexpected.exception';
```
lmarrai-sw: Can we use `@ukef/...` imports please?

test/support/app.ts
```diff
@@ -1,20 +1,33 @@
 import { Test, TestingModule } from '@nestjs/testing';
 import { App as AppUnderTest } from '@ukef/app';
 import { MainModule } from '@ukef/main.module';
+import GraphClientService from '@ukef/modules/graph-client/graph-client.service';
+
+import { MockGraphClientService } from './mocks/graph-client.service.mock';
 
 export class App extends AppUnderTest {
-  static async create(): Promise<App> {
+  mockGraphClientService: MockGraphClientService;
```
lmarrai-sw: Is this used?

test/support/app.ts
```diff
@@ -1,20 +1,33 @@
...
 
     await nestApp.init();
 
-    return app;
+    return { app, mockGraphClientService };
+  }
+
+  getGraphClientServiceMock(): MockGraphClientService {
+    return this.mockGraphClientService;
   }
```
lmarrai-sw: is this used?

test/support/api.ts
```diff
@@ -44,6 +46,10 @@ export class Api {
     return this.app.destroy();
   }
 
+  getGraphClientServiceMock(): MockGraphClientService {
+    return this.app.getGraphClientServiceMock();
+  }
```
lmarrai-sw: Is this used?

test/support/api.ts
```diff
@@ -54,3 +60,8 @@ export class Api {
     return { [strategy]: apiKey };
   }
 }
+
+interface CreateApiResponse {
+  api: Api;
+  mockGraphClientService: MockGraphClientService;
+}
```
lmarrai-sw: I think `Response` in the interface name here is misleading - it makes it sounds like we're talking about a HTTP response (especially with `Api` just before it)

test/site/get-site-status-by-exporter-name.api-test.ts
```diff
@@ -0,0 +1,129 @@
...
+    graphGetSiteStatusResponseDto,
+  } = new getSiteStatusByExporterNameGenerator(valueGenerator).generate({
+    numberToGenerate: 1,
+  });
+
+  let api: Api;
+  let mockGraphClientService: MockGraphClientService;
+
+  beforeAll(async () => {
+    ({ api, mockGraphClientService } = await Api.create());
```
lmarrai-sw, line 24: Do we need the outermost brackets?

test/site/get-site-status-by-exporter-name.api-test.ts
```diff
@@ -0,0 +1,129 @@
...
+    givenRequestWouldOtherwiseSucceed: () => {
+      mockGraphClientService
+        .mockSuccessfulGraphApiCallWithPath(path)
+        .mockSuccessfulExpandCallWithExpandString(expand)
+        .mockSuccessfulFilterCallWithFilterString(filter);
+    },
+    givenGraphServiceCallWillThrowError: (error: Error) => {
+      mockGraphClientService.mockUnsuccessfulGraphGetCall(error);
+    },
+    makeRequest: () => api.get(`/api/v1/sites?exporterName=${siteStatusByExporterNameQueryDto.exporterName}`),
```
lmarrai-sw: We use this string interpolation a few times in this file, can we make a constant or function for it please?

test/site/get-site-status-by-exporter-name.api-test.ts
```diff
@@ -0,0 +1,129 @@
...
+    mockGraphClientService
+      .mockSuccessfulGraphApiCallWithPath(path)
+      .mockSuccessfulExpandCallWithExpandString(expand)
+      .mockSuccessfulFilterCallWithFilterString(filter)
+      .mockSuccessfulGraphGetCall({ value: [] });
+
+    const { status, body } = await api.get(`/api/v1/sites?exporterName=${siteStatusByExporterNameQueryDto.exporterName}`);
+
+    expect(status).toBe(404);
+    expect(body).toStrictEqual({});
```
lmarrai-sw: Why is the response body empty in this case? Do we not have a "Not found" / "Not Found" message for the user?

test/site/get-site-status-by-exporter-name.api-test.ts
```diff
@@ -0,0 +1,129 @@
...
+  });
+
+  it('returns a 400 with validation rules if request does not meet validation rules', async () => {
```
lmarrai-sw: "does not meet validation rules" is a bit vague, and we're only testing one of the validation rules. Can we please test all of them?

test/site/get-site-status-by-exporter-name.api-test.ts
```diff
@@ -0,0 +1,129 @@
...
+    {
+      siteStatus: 'Provisioning',
+      expectedStatusCode: 202,
+    },
+    {
+      siteStatus: 'Failed',
+      expectedStatusCode: 424,
+    },
+  ];
+  it.each(statusCodeTestInputs)('returns $expectedStatusCode if graph replies with $siteStatus', async ({ siteStatus, expectedStatusCode }) => {
```
lmarrai-sw: "if graph replies with" seems a bit unclear to me, can we make it more specific please?

test/support/mocks/graph-client.service.mock.ts
```diff
@@ -0,0 +1,57 @@
+import { Client, GraphRequest } from '@microsoft/microsoft-graph-client';
+import { when } from 'jest-when';
+
+class MockClient {
+  api: jest.Mock;
+  constructor() {
+    this.api = jest.fn();
+  }
```
lmarrai-sw, line 8: Can we have a blank line between the fields and the constructor for all the classes in this file please?

test/support/mocks/graph-client.service.mock.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+class MockRequest {
+  get: jest.Mock<any, any, any>;
+  filter: jest.Mock<any, any, any>;
+  expand: jest.Mock<any, any, any>;
+  constructor() {
+    this.get = jest.fn();
+
```
lmarrai-sw: Is there a reason for this blank line?

test/support/mocks/graph-client.service.mock.ts
```diff
@@ -0,0 +1,57 @@
...
+export class MockGraphClientService {
+  client: Client;
+  request: GraphRequest;
+  constructor() {
+    this.client = new MockClient() as unknown as Client;
+    this.request = new MockRequest() as unknown as GraphRequest;
+  }
+
+  getApiRequestObject() {
```
lmarrai-sw: Can we put return types on all the (non-constructor) functions in this file please?

test/support/mocks/graph-client.service.mock.ts
```diff
@@ -0,0 +1,57 @@
...
+    when(this.request.filter).calledWith(filterString).mockReturnValueOnce(this.request);
+    return this;
+  }
+
+  mockSuccessfulExpandCallWithExpandString(expandString: string) {
+    when(this.request.expand).calledWith(expandString).mockReturnValueOnce(this.request);
+    return this;
+  }
+
+  mockSuccessfulGraphGetCall(response) {
```
lmarrai-sw: Can we have a type for the input here please?

test/support/mocks/graph-client.service.mock.ts
```diff
@@ -0,0 +1,57 @@
...
+    when(this.request.get).calledWith().mockReturnValueOnce(response);
+  }
+
+  mockUnsuccessfulGraphGetCall(error: unknown) {
+    when(this.request.get).calledWith().mockRejectedValueOnce(error);
+  }
```
lmarrai-sw: is it correct that the successful mock uses `mockReturnValueOnce` (i.e. doesn't return a promise) but the unsuccessful mock uses `mockRejectedValueOnce` (i.e. does return a promise)? Should the successful one be `mockResolvedValueOnce`?

test/support/mocks/graph-client.service.mock.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  getApiRequestObject() {
+    return this.request;
+  }
+
+  mockSuccessfulGraphApiCallWithPath(path: string) {
+    when(this.client.api).calledWith(path).mockReturnValueOnce(this.request);
+    return this;
+  }
```
lmarrai-sw: Not an issue for this PR, but to look out for on future tickets: since `client.api` is always returning the same `request` object, we're going to get conflicting mocking/assertions when we use this for endpoints that make more than one call to Microsoft Graph. We'll need to change this as soon as we have endpoints like that

src/modules/graph/graph.service.test.ts
```diff
@@ -24,84 +25,92 @@ describe('GraphService', () => {
...
-
-      expect(mockRequest.expand).toHaveBeenCalledTimes(0);
-    });
-
-    it('chains input parameters as api methods and returns the response', async () => {
-      mockSuccessfulGraphApiCall();
-      mockSuccessfulGraphGetCall();
-      when(mockRequest.filter).calledWith(filterStr).mockReturnValueOnce(mockRequest);
-      when(mockRequest.expand).calledWith(expandStr).mockReturnValueOnce(mockRequest);
+    it('calls all graph client methods on a graph serviec get request with multiple parameters and returns the response', async () => {
```
lmarrai-sw: small typo in `serviec` here :)

src/modules/graph/graph.service.test.ts
```diff
@@ -24,84 +25,92 @@ describe('GraphService', () => {
...
+    getCalled = false,
+  }: {
+    apiCalled?: boolean;
+    filterCalled?: boolean;
+    expandCalled?: boolean;
+    getCalled?: boolean;
+  }) => {
+    apiCalled
+      ? (expect(mockGraphClientService.client.api).toHaveBeenCalledTimes(1), expect(mockGraphClientService.client.api).toHaveBeenCalledWith(path))
+      : expect(mockGraphClientService.client.api).toHaveBeenCalledTimes(0);
```
lmarrai-sw: Personally I don't like the "evaluate an expression that evaluates two void functions in a tuple and then throw away the tuple" trick that we're doing here to make this syntax work, I think it's too surprising to be executing void functions in this way. I'd prefer us to just use if statements

Or we could refactor this function to look like
```
const expectations = getCallExpectations({ apiCalled, filterCalled, expandCalled, getCalled });
expectations.forEach((expectation) => expectation());
```
where `getCallExpectations` looks something like
```
const apiCallExpectations = apiCalled
  ? [ () => expect(mockGraphClientService.client.api).toHaveBeenCalledTimes(1), () => expect(mockGraphClientService.client.api).toHaveBeenCalledWith(path)]
  : [ () => expect(mockGraphClientService.client.api).toHaveBeenCalledTimes(0)]
// etc.
```

lmarrai-sw (commented):
Thanks for this @AlexBramhill, some really great re-usable tet stuff here

src/modules/graph/graph.service.ts
AlexBramhill: The reason we moved away from the field in the first place was due to mocking in tests -- As we've developed a way to do so without requiring the getter, I moved it back

test/support/app.ts
AlexBramhill: Ah -- No sorry! This was an approach that I later decided against, instead choosing to pass out `mockGraphClientService` in the return value when creating the test app

test/support/mocks/graph-client.service.mock.ts
AlexBramhill: Ah -- I might add a comment or just remove. I was seperating out the final method vs the chaining methods

test/support/mocks/graph-client.service.mock.ts
AlexBramhill: Good catch!

AlexBramhill (commented):
Ill pick these up first thing

src/modules/graph/graph.module.ts
AlexBramhill: Ah -- That was me mucking with creating test modules...Reverted

test/docs/get-docs.api-test.ts
AlexBramhill: As we already declare the `api` variable as a `let`, we cannot destructure onto it without the parentheses

test/docs/get-docs.api-test.ts
AlexBramhill: Same answer as above :)

test/support/mocks/graph-client.service.mock.ts
AlexBramhill: I've changed this to be a generic method as this can return potentially anything from Microsoft Graph

src/modules/site/site.service.test.ts
AlexBramhill: Nice nice -- I've updated this. Needed to update the generator as a result too :)

src/modules/graph/graph.service.ts
AlexBramhill: There was no need for the getter (I think). It was originally a field. It's only consumed by graph.service. It was changed to a getter to get it easier to mock -- but since we've found a better way, I changed it back to a field

test/docs/__snapshots__/get-docs.api-test.ts.snap
AlexBramhill: I thought for tfs, if we were simply passing through the enum, we would do so as a string?

test/docs/__snapshots__/get-docs.api-test.ts.snap
avaitonis: Site ids have 8 digits.
In Dev and Prod environments they are matching pattern 0070\d{4}. Maybe 00701234 can be an example/

test/docs/__snapshots__/get-docs.api-test.ts.snap
lmarrai-sw: (Alex and I had a chat about this: for GET responses in TFS, we didn't use an enum because we couldn't guarantee that the data in ACBS would follow that enum. For estore Alex checked that we can)

src/modules/site/site.controller.ts
```diff
@@ -36,27 +35,19 @@ export class SiteController {
...
+    const getSiteStatusByExporterNameResponse = await this.service.getSiteStatusByExporterName(query.exporterName);
+    if (getSiteStatusByExporterNameResponse.status === 'Failed') {
+      res.status(HttpStatusCode.FailedDependency).json(getSiteStatusByExporterNameResponse);
+      return;
     }
+    if (getSiteStatusByExporterNameResponse.status === 'Created') {
+      res.status(HttpStatusCode.Ok).json(getSiteStatusByExporterNameResponse);
+      return;
+    }
+    if (getSiteStatusByExporterNameResponse.status === 'Provisioning') {
```
lmarrai-sw: Can we use the enum instead of referring to the strings directly here please? (I think it was just missed when you added the enum)

lmarrai-sw (changes requested):

lmarrai-sw (approved):

PR #44, opened 2023-06-17
feat(APIM-142): migrate the POST `/term/facility` endpoint

IgnatG:
### Introduction
We need to add facilityIds to estore term store. The endpoint is POST `/terms/facilities`

### Resolution
Added a new endpoint that adds the facilityId to TFIS_TERM_STORE

### TODO
Add API tests


.env.sample
```diff
@@ -16,6 +16,11 @@ API_KEY=
 # Microsoft Graph
 GRAPH_BASE_URL=
 
+SHAREPOINT_MAIN_SITE_NAME=
+SHAREPOINT_TFIS_SITE_NAME=
+SHAREPOINT_TFIS_LIST_ID=
+SHAREPOINT_TFIS_TERM_STORE=
```
lmarrai-sw: We'll also need to add these to `docker-compose.yml`, `.github/workflows/development.yml`, and `.github/workflows/test.yml`

README.md
```diff
@@ -140,6 +140,7 @@ this.logger.error({ id: 'your message here' }, 'context-name');
 # log
 this.logger.log({ id: 'your message here' }, 'context-name');
 
+
```
lmarrai-sw: This change was probably an accident :)

src/modules/terms/terms.service.test.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  describe('postFacilityToTermStore', () => {
+    it('should call the graphService post method with the correct parameters and return a success message', async () => {
+      when(mockGraphServicePost).calledWith(expect.any(Object)).mockResolvedValue(undefined);
+
+      const result = await service.postFacilityToTermStore(facilityId);
+
+      expect(graphService.post).toHaveBeenCalled();
+      expect(result).toEqual({ message: RESPONSE.FACILITY_TERM_CREATED });
+    });
```
lmarrai-sw: > with the correct parameters

I don't think this test checks that the parameters used by the service are correct, since we don't check the parameters that `graphService.post` are called with

src/modules/terms/terms.controller.ts
```diff
@@ -0,0 +1,35 @@
...
+import { RESPONSE } from '@ukef/constants';
+
+import { CreateFacilityTermRequestItem } from './dto/create-facility-term-request.dto';
+import { TermsService } from './terms.service';
+
+@Controller('terms')
+export class TermsController {
+  constructor(private readonly service: TermsService) {}
+
+  @Post('facilities')
```
lmarrai-sw, line 14: It'd be great if we could write API tests for this endpoint too. Alex is still working out how to do these on his ticket so it'd make sense to do API tests for this endpoint as a separate ticket to avoid the overlap. I've raised APIM-461 for that

src/constants/response.constant.ts
```diff
@@ -0,0 +1,6 @@
+export const RESPONSE = {
+  FACILITY_TERM_CREATED: 'Facility Term created',
+  FACILITY_TERM_NOT_CREATED: 'Facility Term not created',
```
lmarrai-sw: I don't think this value is used

src/constants/response.constant.ts
```diff
@@ -0,0 +1,6 @@
...
+  FACILITY_TERM_EXISTS: 'Facility Term already exists',
+  INTERNAL_SERVER_ERROR: 'Internal Server Error',
```
lmarrai-sw: These 2 values are only used by tests so they might be better in the `test/support` folder

src/modules/terms/dto/create-facility-term-response.dto.ts
```diff
@@ -0,0 +1,13 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { RESPONSE } from '@ukef/constants';
+
+export class CreateDealResponse {
+  @ApiProperty({
+    example: RESPONSE.FACILITY_TERM_CREATED,
+  })
+  message: string;
```
lmarrai-sw: Maybe we could make it clearer (in the swagger specification and in the type system) that this value will always be `Facility Term created`

For the type system, we could do
```
message = 'FacilityTermCreated' as const;
```
and for the swagger specification I think we could document it as an enum with only 1 possible value

src/modules/terms/dto/create-facility-term-response.dto.ts
```diff
@@ -0,0 +1,13 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { RESPONSE } from '@ukef/constants';
+
+export class CreateDealResponse {
```
lmarrai-sw: I think the class name needs renaming

src/modules/terms/dto/create-facility-term-response.dto.ts
```diff
@@ -0,0 +1,13 @@
...
+  @ApiProperty({
+    example: RESPONSE.FACILITY_TERM_CREATED,
+  })
+  message: string;
+
+  constructor(message: string) {
+    this.message = message;
+  }
```
lmarrai-sw: We decided in TFS that we didn't need the constructor (although some of the older code in TFS still has it)

src/modules/terms/terms.controller.ts
```diff
@@ -0,0 +1,35 @@
...
+  @ApiOperation({
+    summary: 'Add a new facility ID to term store.',
+  })
+  @ApiBody({
+    type: CreateFacilityTermRequestItem,
+  })
+  @ApiCreatedResponse({
+    description: RESPONSE.FACILITY_TERM_CREATED,
+    type: CreateFacilityTermRequestItem,
```
lmarrai-sw: This should be the response class instead

src/modules/terms/terms.controller.ts
```diff
@@ -0,0 +1,35 @@
...
+  @ApiNotFoundResponse({
+    description: 'The facility ID was not found.',
+  })
+  @ApiBadRequestResponse({
+    description: 'Bad request.',
+  })
+  @ApiInternalServerErrorResponse({
+    description: 'An internal server error has occurred.',
+  })
+  postFacilityToTermStore(@Body() term: CreateFacilityTermRequestItem): Promise<CreateFacilityTermRequestItem> {
```
lmarrai-sw: In 3 amigos last week, we decided that we want to make the request body an array containing the object, instead of just the object, to be consistent with the other POST endpoints in TFS and eStore. Sorry, I realise I didn't click "save" on the Jira ticket to update it last week so that part of the spec only appeared just now!

src/modules/terms/dto/create-facility-term-request.dto.ts
```diff
@@ -0,0 +1,17 @@
...
+    example: EXAMPLES.FACILITY_ID,
+    minLength: 10,
+    maxLength: 10,
+    pattern: UKEFID.MAIN_ID.TEN_DIGIT_REGEX,
+  })
+  readonly id: string;
+
+  constructor(id: string) {
+    this.id = id;
+  }
```
lmarrai-sw: We decided in TFS that we didn't need the constructor (although some of the older code in TFS still has it)

src/modules/terms/terms.controller.ts
lmarrai-sw: (so there's nothing to do in this PR for it!)

lmarrai-sw (changes requested):
I've commented on a few small things but otherwise this is great, thanks @IgnatG

README.md
IgnatG: yup ;)

src/constants/response.constant.ts
IgnatG: good spot - I was using it previously, but reverted the functionality and forgot to take this out.

.env.sample
avaitonis: Added SHAREPOINT_TFIS_TERM_STORE  to these files

README.md
avaitonis: Removed line

src/modules/terms/terms.service.test.ts
avaitonis: Made description more generic and, I hope, correct.

src/constants/response.constant.ts
avaitonis: Removed old constant

src/constants/response.constant.ts
avaitonis: Moved constants to `test/support`

src/modules/terms/dto/create-facility-term-response.dto.ts
avaitonis: I created Enum for response and return it's value. Also it's used in ApiProperty

Tried to add `as const` in couple places, but **it didn't worked**!

src/modules/terms/dto/create-facility-term-response.dto.ts
avaitonis: Changed to CreateTermFacilityResponse to match URL and module name.

src/modules/terms/dto/create-facility-term-response.dto.ts
avaitonis: Removed

src/modules/terms/terms.controller.ts
avaitonis: Changed

src/modules/terms/terms.controller.ts
avaitonis: Changed request to array.

src/modules/terms/dto/create-facility-term-request.dto.ts
avaitonis: Removed constructor.

lmarrai-sw (approved):
Thanks @avaitonis

PR #46, opened 2023-06-19
feat(APIM-450): check file exists and does not exceed max size

oscar-richardson-softwire:
### Introduction

As part of [APIM-138](https://ukef-dtfs.atlassian.net/browse/APIM-138), we need to check if a file exists in DTFS Azure File storage and check that it does not exceed a maximum allowed size. This PR adds this functionality in the form of a temporary endpoint (the endpoint will ultimately be removed but the associated services, methods etc. will remain). Note that this functionality currently uses `shared key` auth with the Azure storage account; I'm still investigating whether the either of the other options, `Azure AD` or `SAS` auth, would be better options, hence this PR is a draft at present.
,
### Resolution

I've added a `dtfs` module which handles sending/receiving requests to/from DTFS file storage using the [Azure SDK](https://github.com/Azure/azure-sdk-for-js), which also handles auth. I've also added a `temporary file check` module containing a `file service` which calls through to a service in the `dtfs` module and contains the logic about checking the file isn't too large and handling if it does exist.


test/support/dtfs-authentication.service.mock.ts
```diff
@@ -0,0 +1,21 @@
+import { DtfsAuthenticationService } from '@ukef/modules/dtfs-authentication/dtfs-authentication.service';
+
+export const getMockDtfsAuthenticationService = (): {
```
AlexBramhill: Might want to have a look how I've done it on my latest branch

src/constants/max-file-size.constant.ts
```diff
@@ -0,0 +1 @@
+export const MAX_FILE_SIZE = 12000000;
```
lmarrai-sw: Can we specify the units in the variable name please?

src/modules/temporary-file-check/dto/get-file-size-request.dto.ts
```diff
@@ -0,0 +1,24 @@
...
+  fileName: string;
+
+  @ValidatedStringApiProperty({
+    description: 'The location or file path in Azure storage from which the file needs to be loaded.',
+    minLength: 1,
+    maxLength: 250,
+    pattern: /^[A-Za-z/-_.()\s\-]+$/,
+    example: EXAMPLES.FILE_LOCATION_PATH,
+  })
+  fileLocationPath: string;
```
lmarrai-sw: Is there a reason we're using different `pattern`s than Mulesoft has documented?

src/constants/max-file-size.constant.ts
lmarrai-sw: Also, I think it'll be worth a comment that says why this is the max file size

.env.sample
```diff
@@ -27,4 +27,9 @@ GRAPH_AUTHENTICATION_CLIENT_SECRET=
 # SHAREPOINT
 SHAREPOINT_MAIN_SITE_NAME=
 SHAREPOINT_TFIS_SITE_NAME=
-SHAREPOINT_TFIS_LIST_ID=
\ No newline at end of file
+SHAREPOINT_TFIS_LIST_ID=
+
+# DTFS STORAGE
```
abhi-markan: Rename to `STORAGE` please so we can use this for other applications going forward.

src/modules/dtfs-storage/dtfs-storage-file.service.ts
```diff
@@ -0,0 +1,34 @@
...
+    @Inject(DtfsConfig.KEY)
+    { baseUrl, accountName, accountKey }: ConfigType<typeof DtfsConfig>,
+  ) {
+    this.baseUrl = baseUrl;
+    this.storageSharedKeyCredential = new StorageSharedKeyCredential(accountName, accountKey);
+  }
+
+  async getFileProperties(fileName: string, fileLocationPath: string): Promise<FileGetPropertiesResponse> {
+    const url = `${this.baseUrl}/${fileLocationPath}/${fileName}`;
+    const options = { fileRequestIntent: 'backup' };
```
lmarrai-sw: This is a bit surprising. From doing some googling (e.g. https://github.com/Azure/azure-sdk-for-js/pull/25005/commits/27e24c66927ae703e2d2176a4ad0ff0ab15d48c8 and https://github.com/Azure/azure-sdk-for-js/pull/25005#discussion_r1124841390), it sounds like `'backup'` is the only valid option but we have to provide it - is that what you found too?

It might be worth a comment in our own code about that if so

src/config/dtfs-storage.config.ts
```diff
@@ -0,0 +1,15 @@
+import { registerAs } from '@nestjs/config';
+
+export interface DtfsStorageConfig {
```
abhi-markan, line 3: Camel case instead of Pascal case please?

src/modules/dtfs-storage/dtfs-storage-file.service.ts
```diff
@@ -0,0 +1,34 @@
...
+    const shareFileClient = new ShareFileClient(url, this.storageSharedKeyCredential, options);
+
+    return await shareFileClient.getProperties().catch((error) => {
+      if (error.statusCode === 404) {
+        throw new DtfsStorageFileNotFoundException(`File ${fileLocationPath}/${fileName} was not found in DTFS.`, error);
+      }
+      throw new DtfsStorageException(`Failed to get properties for file ${fileLocationPath}/${fileName}.`, error);
+    });
```
lmarrai-sw: You probably just want to `return` here instead of `return await` (unless you need the exception to be thrown _here_ rather than where `getFileProperties` is called)

src/constants/examples.constant.ts
```diff
@@ -1,4 +1,9 @@
+import { MAX_FILE_SIZE } from './max-file-size.constant';
+
 export const EXAMPLES = {
   SITE_ID: '12345678',
   SITE_STATUS_CODE: 'Created',
+  FILE_SIZE: MAX_FILE_SIZE / 2,
```
abhi-markan: Will be good to a comment regarding division by 2 please.

src/modules/dtfs-storage/dtfs-storage-file.service.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
```
lmarrai-sw: Can we break this into slightly smaller functions please?

src/modules/temporary-file-check/file.service.get-file-size.test.ts
```diff
@@ -0,0 +1,50 @@
+import { FileGetPropertiesResponse } from '@azure/storage-file-share';
+import { BadRequestException } from '@nestjs/common';
+import { MAX_FILE_SIZE } from '@ukef/constants';
+import { RandomValueGenerator } from '@ukef-test/support/generator/random-value-generator';
+import { when } from 'jest-when';
+
+import { DtfsStorageFileService } from '../dtfs-storage/dtfs-storage-file.service';
```
lmarrai-sw: Can we use `@ukef/modules` import please?

src/constants/max-file-size.constant.ts
abhi-markan: Can we rename it to `MAX_FILE_SIZE_BYTES` and set the value to `12582912` so we are using `1024` as a base instead of `1000` for improved accuracy.

src/modules/temporary-file-check/file.service.ts
```diff
@@ -0,0 +1,25 @@
...
+import { GetFileSizeResponse } from './dto/get-file-size-response.dto';
+
+@Injectable()
+export class FileService {
+  constructor(private readonly dtfsStorageFileService: DtfsStorageFileService) {}
+
+  async getFileSize(fileToCheck: GetFileSizeRequestItem): Promise<GetFileSizeResponse> {
+    const { fileName, fileLocationPath } = fileToCheck;
+    const fileSize = await this.dtfsStorageFileService.getFileProperties(fileName, fileLocationPath).then((response) => {
+      return response.contentLength as number;
```
lmarrai-sw: I'm surprised we need `as number` here - does the package's type system not think it's a `number`?

src/modules/temporary-file-check/file.service.ts
```diff
@@ -0,0 +1,25 @@
...
+    });
+    if (fileSize > MAX_FILE_SIZE) {
+      throw new BadRequestException('Bad request', 'The file exceeds the maximum allowed size.');
+    }
```
lmarrai-sw: It's not a huge deal immediately, but we're probably going to want to move this out of the current function when we actually work on this endpoint (because it's surprising for a function called `getFileSize` to be throwing a BadRequestException about the file being too large). We can worry about that in the next ticket though

src/modules/temporary-file-check/file.service.ts
```diff
@@ -0,0 +1,25 @@
...
+import { GetFileSizeRequestItem } from './dto/get-file-size-request.dto';
+import { GetFileSizeResponse } from './dto/get-file-size-response.dto';
+
+@Injectable()
+export class FileService {
+  constructor(private readonly dtfsStorageFileService: DtfsStorageFileService) {}
+
+  async getFileSize(fileToCheck: GetFileSizeRequestItem): Promise<GetFileSizeResponse> {
+    const { fileName, fileLocationPath } = fileToCheck;
+    const fileSize = await this.dtfsStorageFileService.getFileProperties(fileName, fileLocationPath).then((response) => {
```
lmarrai-sw: `fileSizeInBytes` would be a bit more helpful as a variable name

src/modules/temporary-file-check/file.service.ts
```diff
@@ -0,0 +1,25 @@
...
+      return response.contentLength as number;
+    });
+    if (fileSize > MAX_FILE_SIZE) {
+      throw new BadRequestException('Bad request', 'The file exceeds the maximum allowed size.');
```
abhi-markan: Might a good idea to print `MAX_FILE_SIZE` in the error.

src/modules/dtfs-storage/dtfs-storage-exception-transform.interceptor.ts
```diff
@@ -0,0 +1,21 @@
+import { BadRequestException, CallHandler, ExecutionContext, Injectable, NestInterceptor } from '@nestjs/common';
+import { catchError, Observable, throwError } from 'rxjs';
+
+import { DtfsStorageFileNotFoundException } from './exception/dtfs-storage-file-not-found.exception';
+
+@Injectable()
+export class DtfsStorageExceptionTransformInterceptor implements NestInterceptor {
```
lmarrai-sw, line 7: Nice idea with using an interceptor here 👏

abhi-markan (changes requested):
Few comments please.

src/modules/estore.module.ts
```diff
@@ -1,13 +1,21 @@
...
 @Module({
-  imports: [AuthModule, GraphClientModule, GraphModule, SiteModule],
-  providers: [],
+  imports: [AuthModule, GraphClientModule, GraphModule, SiteModule, TemporaryFileCheckModule],
+  providers: [
+    {
+      provide: APP_INTERCEPTOR,
+      useClass: DtfsStorageExceptionTransformInterceptor,
+    },
+  ],
```
lmarrai-sw: We're registering the interceptor globally here, did you consider registering the interceptor per-endpoint instead?

lmarrai-sw (changes requested):
Great work here @oscar-richardson-softwire, let me know when you want to discuss the auth approach when you have time to

test/support/dtfs-authentication.service.mock.ts
oscar-richardson-softwire: Thanks, didn't need this mock in the end as ended up taking a different approach

.env.sample
oscar-richardson-softwire: Do you mean just the environment variables, or everywhere (i.e., the name of the module/service should change from `DtfsStorage` to just `Storage` too?

src/config/dtfs-storage.config.ts
oscar-richardson-softwire: In both estore and tfs we've been using Pascal case for interfaces, types and classes

src/constants/examples.constant.ts
oscar-richardson-softwire: The division by 2 was pretty arbitrary, just wanted some amount that was less than or equal to the `MAX_FILE_SIZE`, but I've realised I may as well just use `MAX_FILE_SIZE`

src/constants/max-file-size.constant.ts
oscar-richardson-softwire: Changed

src/modules/temporary-file-check/file.service.ts
oscar-richardson-softwire: Changed to `The file exceeds the maximum allowed size of ${MAX_FILE_SIZE_BYTES} bytes.`

src/modules/temporary-file-check/dto/get-file-size-request.dto.ts
oscar-richardson-softwire: I initially went off the `pattern`s in the ticket, but found that these didn't work for any of the real files in the dev storage account, so I adapted them to make them work. I meant to raise this when I put up this PR yesterday but didn't get the chance, sorry! I've seen that the ticket has since been updated (and I've also double checked these regexes against the ones in Mulesoft), so I've changed them to match Mulesoft

src/modules/dtfs-storage/dtfs-storage-file.service.ts
oscar-richardson-softwire: That was my understanding as well, but I hadn't tried omitting it, and having just tried that now, it still works. So I'm happy to leave it out altogether?

src/modules/dtfs-storage/dtfs-storage-file.service.ts
oscar-richardson-softwire: Changed

src/modules/dtfs-storage/dtfs-storage-file.service.ts
oscar-richardson-softwire: Changed

src/modules/temporary-file-check/file.service.get-file-size.test.ts
oscar-richardson-softwire: Changed

src/modules/temporary-file-check/file.service.ts
oscar-richardson-softwire: You're right, it's not needed, so I've removed it

src/modules/temporary-file-check/file.service.ts
oscar-richardson-softwire: Changed

src/modules/estore.module.ts
oscar-richardson-softwire: I realised I should've registered it per-endpoint yesterday after I raised this PR. I've changed it

src/constants/max-file-size.constant.ts
oscar-richardson-softwire: What is the reason? That this is the maximum file size we support in DTFS, or in Sharepoint?

oscar-richardson-softwire: Also changed variable name to `MAX_FILE_SIZE_BYTES`

src/constants/examples.constant.ts
```diff
@@ -1,6 +1,10 @@
 import { ENUMS } from './enum.constant';
+import { MAX_FILE_SIZE_BYTES } from './max-file-size-bytes.constant';
 
 export const EXAMPLES = {
   SITE_ID: '12345678',
   SITE_STATUS_CODE: ENUMS.SITE_STATUSES.CREATED,
+  FILE_SIZE: MAX_FILE_SIZE_BYTES,
+  FILE_NAME: 'document.txt',
+  FILE_LOCATION_PATH: 'folder/subfolder',
```
oscar-richardson-softwire: As the Swagger page auto-fills with these values, it would be nice for testing purposes to use the real file name and path of a file in DFTS dev storage, but does this pose any kind of privacy risk? @abhi-markan

src/constants/max-file-size-bytes.constant.ts
```diff
@@ -0,0 +1 @@
+export const MAX_FILE_SIZE_BYTES = 12582912;
```
AlexBramhill: Do we want this file named like this? It's only used for dtfs -- would `dtfs-max-file-size-bytes` be better?

src/constants/examples.constant.ts
```diff
@@ -1,6 +1,10 @@
 import { ENUMS } from './enum.constant';
+import { MAX_FILE_SIZE_BYTES } from './max-file-size-bytes.constant';
 
 export const EXAMPLES = {
   SITE_ID: '12345678',
   SITE_STATUS_CODE: ENUMS.SITE_STATUSES.CREATED,
+  FILE_SIZE: MAX_FILE_SIZE_BYTES,
```
AlexBramhill, line 9: `FILE_SIZE_BYTES`?

src/modules/dtfs-storage/dtfs-storage-file.service.ts
```diff
@@ -0,0 +1,40 @@
...
+    this.storageSharedKeyCredential = new StorageSharedKeyCredential(accountName, accountKey);
+  }
+
+  getFileProperties(fileName: string, fileLocationPath: string): Promise<FileGetPropertiesResponse> {
+    const shareFileClient = this.getShareFileClient(fileName, fileLocationPath);
+
+    return shareFileClient.getProperties().catch((error) => this.handleGetFilePropertiesError(error, fileName, fileLocationPath));
+  }
+
+  private getShareFileClient(fileName: string, fileLocationPath: string): ShareFileClient {
```
AlexBramhill, line 28: As mentioned in a huddle -- We might want to abstract this out into a `dtfs-storage-client` or similar to allow for similar unit testing to `graph-client-service`... but this can wait until we need it (later on in this story)

src/modules/temporary-file-check/file.service.ts
```diff
@@ -0,0 +1,25 @@
+import { BadRequestException, Injectable } from '@nestjs/common';
+import { MAX_FILE_SIZE_BYTES } from '@ukef/constants';
+import { DtfsStorageFileService } from '@ukef/modules/dtfs-storage/dtfs-storage-file.service';
+
+import { GetFileSizeRequestItem } from './dto/get-file-size-request.dto';
+import { GetFileSizeResponse } from './dto/get-file-size-response.dto';
+
+@Injectable()
+export class FileService {
```
AlexBramhill, line 9: Out of interest is this temporary ie going to be removed later, or temporary ie checking for temporary files? Assuming the former with the naming of this service compared to the folder

src/constants/examples.constant.ts
oscar-richardson-softwire: Personally I think it's most useful if the name of the example matches the name of the field it is an example for, which is `fileSize`

src/modules/temporary-file-check/file.service.ts
oscar-richardson-softwire: The former, as in going to be removed later

src/constants/max-file-size.constant.ts
lmarrai-sw: It's the maximum file size we support in DTFS :)

src/modules/dtfs-storage/dtfs-storage-file.service.ts
lmarrai-sw: All of the sample docs I can see don't mention setting it either. Do you remember why we first thought we should set it?

lmarrai-sw (commented):
Thanks for the update Oscar :)

src/modules/temporary-file-check/dto/get-file-size-request.dto.ts
lmarrai-sw: Yep sorry for the ticket changing, when I made the comment above I realised I had put the wrong patterns on the ticket!

src/constants/max-file-size.constant.ts
oscar-richardson-softwire: Added comment

src/constants/max-file-size-bytes.constant.ts
oscar-richardson-softwire: Agreed, renamed

src/modules/dtfs-storage/dtfs-storage-file.service.ts
oscar-richardson-softwire: It was when I was looking at the [ShareClientConfig interface docs](https://learn.microsoft.com/en-us/javascript/api/@azure/storage-file-share/shareclientconfig?view=azure-node-latest) to see what could be passed in the (optional) options parameter to the `ShareFileClient`. The description makes it sound like it's required

lmarrai-sw: Oscar and I spoke about this on Slack. The documentation that Oscar linked makes it sound like it's required for "Azure File OAuth over REST", which is referring to the Azure AD auth method as far as we can tell. Since we are using the account key auth method instead, we don't think it applies to us, and we've tested that it works without the option. Therefore we're going to remove it

lmarrai-sw (dismissed):

.env.sample
abhi-markan: Sorry for the confusion I meant rename `DTFS STORAGE` to `STORAGE` please?

src/constants/examples.constant.ts
abhi-markan: Can these be loaded from .env?

.env.sample
oscar-richardson-softwire: Just to clarify my question, I mean do you want that change just in that place (i.e. just line 32 in the .env file), or everywhere in the codebase (i.e., we shouldn't refer to `Dtfs Storage` in any class/file names, just `Storage`)?

src/constants/examples.constant.ts
oscar-richardson-softwire: Great idea, I've made that change

.env.sample
abhi-markan: Just in that place please `.env.sample` please.

oscar-richardson-softwire: Done

oscar-richardson-softwire:
Closed as draft PR, changes merged in under different branch/PR

PR #57, opened 2023-06-21
feat(APIM-133): create a site id via the mdm /numbers endpoint

lmarrai-sw:
## Introduction

For the `POST /sites` endpoint, we need to be able to generate a site ID for the new site via the MDM `POST /numbers` endpoint.

## Resolution

@francescastocco and I have:
- added the `http` module that we created in TFS to allow us to make http requests with logging built-in
- created a `mdm` module that can be used to send a request to `POST /numbers`, and will wrap any errors in an MdmException
- created `MdmConfig` that will parse new environment variables to configure the requests to MDM in the `mdm` module
- added a method to `SiteService` to create and return the site ID
- added unit tests for the above


src/config/mdm.config.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export default registerAs(
+  KEY,
+  (): MdmConfig => ({
+    baseUrl: process.env.MDM_BASE_URL,
+    apiKeyHeaderName: process.env.MDM_API_KEY_HEADER_NAME,
+    apiKeyHeaderValue: process.env.MDM_API_KEY_HEADER_VALUE,
+    // TODO APIM-133: what about 0?
+    maxRedirects: parseInt(process.env.MDM_MAX_REDIRECTS) || 5,
+    timeout: parseInt(process.env.MDM_TIMEOUT) || 30000,
```
lmarrai-sw: We use `parseInt || <default>` to parse numeric config in TFS as well, but this means we can never set the value to 0! I'm going to raise a low-priority ticket to change this across estore and TFS

src/modules/site/site.service.ts
```diff
@@ -31,4 +34,23 @@ export class SiteService {
     const { URL: siteId, Sitestatus: status } = data.value[0].fields;
     return { siteId, status };
   }
+
+  async createSiteId(): Promise<string> {
+    const requestToCreateSiteId: MdmCreateNumbersRequest = this.buildRequestToCreateSiteId();
+    // TODO APIM-133: Do we want to wrap any errors in a new error?
+    const [{ maskedId: createdSiteId }] = await this.mdmService.createNumbers(requestToCreateSiteId);
```
lmarrai-sw: @avaitonis it might be best to answer this with knowledge of your POST /sites ticket. Errors from `mdmService.createNumbers` will be of type `MdmException` and will have the message `Failed to create numbers in MDM.`.

Do we want to re-wrap those errors into a message that says something more like `Failed to create a site ID` _here_? Or do we want to do anything like that in your ticket instead? (Personally, I think it'll make more sense in your ticket)

src/modules/site/site.service.ts
```diff
@@ -31,4 +34,23 @@ export class SiteService {
...
+  private buildRequestToCreateSiteId(): MdmCreateNumbersRequest {
+    // TODO APIM-133: Should we use the config for app.name for this, or are they separate concepts?
+    const applicationName = 'Estore';
+    return [
+      {
+        numberTypeId: 6,
+        createdBy: applicationName,
+        requestingSystem: applicationName,
+      },
+    ];
```
lmarrai-sw: @avaitonis do you know much about the significance of `createdBy` and `requestingSystem` here? I'm wondering if we should link these to the `app.name` config property (which is currently `estore` instead of `Estore`)

lmarrai-sw (commented):
Highlighting the TODOs here:

src/modules/site/site.service.ts
avaitonis: Thanks, happy todo it my ticket

src/modules/site/site.service.ts
avaitonis: `createdBy` and `requestingSystem` are not very important. It is fine to use app.name and have lower case `estore`, leaving `Estore` is also fine.

src/config/mdm.config.ts
```diff
@@ -0,0 +1,23 @@
...
+  timeout: number;
+}
+
+export default registerAs(
+  KEY,
+  (): MdmConfig => ({
+    baseUrl: process.env.MDM_BASE_URL,
+    apiKeyHeaderName: process.env.MDM_API_KEY_HEADER_NAME,
+    apiKeyHeaderValue: process.env.MDM_API_KEY_HEADER_VALUE,
+    // TODO APIM-133: what about 0?
```
AlexBramhill: A reminder to delete this comment after raising ticket

src/modules/mdm/dto/mdm-create-numbers-request.dto.ts
```diff
@@ -0,0 +1,7 @@
+export type MdmCreateNumbersRequest = MdmCreateNumbersRequestItem[];
+
+interface MdmCreateNumbersRequestItem {
+  numberTypeId: number;
```
AlexBramhill, line 4: Is this possible to enum? I'm just concerned about the abstractness of this field on the request

src/modules/mdm/mdm.service.test.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  let httpServicePost: jest.Mock;
+  let service: MdmService;
+
+  beforeEach(() => {
+    const httpService = new HttpService();
+    httpServicePost = jest.fn();
+    httpService.post = httpServicePost;
+
+    service = new MdmService(httpService);
```
AlexBramhill, line 23: do we want a resetAllWhenMocks() in here to ensure we never have overspill?

src/modules/mdm/mdm.service.ts
```diff
@@ -0,0 +1,54 @@
...
+
+import { MdmCreateNumbersRequest } from './dto/mdm-create-numbers-request.dto';
+import { MdmCreateNumbersResponse } from './dto/mdm-create-numbers-response.dto';
+import { MdmException } from './exception/mdm.exception';
+
+@Injectable()
+export class MdmService {
+  constructor(private readonly httpService: HttpService) {}
+
+  async createNumbers(numbersToCreate: MdmCreateNumbersRequest): Promise<MdmCreateNumbersResponse> {
```
AlexBramhill, line 14: Is this the pattern were going to use for this service? ie having a method for each request?

AlexBramhill (dismissed):
lgtm

src/modules/mdm/dto/mdm-create-numbers-request.dto.ts
lmarrai-sw: Good question! I'll check

src/modules/mdm/mdm.service.ts
lmarrai-sw: I was thinking we could start with a method per request while there's only 1 or 2 requests and then we could make a change like having separate classes for each endpoint when it gets too big. This is the only MDM request we need to make for the ESTORE endpoints we're migrating, so we shouldn't need to make that change for a long while

src/modules/mdm/mdm.service.test.ts
lmarrai-sw: Good question. In this case:

- We create new `jest` mock functions in the `beforeEach` so there's nothing to reset before the tests.
- The mocks in this file are all local variables so can't affect any tests outside of this file so there's no need to clean them up in an `afterEach` either

src/config/mdm.config.ts
lmarrai-sw: Raised APIM-468

src/modules/site/site.service.ts
lmarrai-sw: I'll leave it as-is then, thank you

src/modules/mdm/dto/mdm-create-numbers-request.dto.ts
lmarrai-sw: The mdm-api code does not use an enum, but the validation says that it does have to be an integer and the maximum is 9. I've checked the dev db and the maximum is actually 8.

@avaitonis do you know why MDM's validation uses a maximum of 9 instead?

avaitonis: Not sure why 9 is max. Maybe to test how DB SP behaves with invalid types.

For estore we will have just one numberTypeId value, not sure if we need info about other types in MDM.

AlexBramhill: Happy to leave as is then -- @lmarrai-sw ill leave you to resolve this comment to ensure you've seen the answer:)

src/modules/mdm/mdm.service.ts
AlexBramhill: That sounds like a good approach -- I'm also keen to hold off splitting if it gets too big!

AlexBramhill (approved):
lgtm

PR #59, opened 2023-06-22
Feat/apim 134/migrate get site exists endpoint

AlexBramhill:
# Intro
This addresses the query parameter requirements for the GET site by exporter name endpoint

# Resolution

Updated regex to match `/^[A-Za-z\d\-._()\s]+$/`
Updated max length to 250
Added tests


src/constants/exporter-name.constant.ts
```diff
@@ -0,0 +1,3 @@
+export const EXPORTER_NAME = {
+  REGEX: /^[\w\-.()\s]+$/,
```
AlexBramhill, line 2: Automatically changed from `/^[A-Za-z\d\-._()\s]+$/` by linter. `\w` - 
Matches any word character (alphanumeric & underscore). Only matches low-ascii characters (no accented or non-roman characters). Equivalent to [A-Za-z0-9_]

lmarrai-sw: Ah exciting! Thank you for explaining

lmarrai-sw (approved):
LGTM thanks @AlexBramhill

PR #61, opened 2023-06-23
feat(APIM-454): migrate POST `/site` endpoint

avaitonis:
### Introduction
Endpoint POST `/sites` is required to create new sites in Sharepoint. API code will create list item in Sharepoint. Power app will pick up list item and start Sharepoint site creation.

### Resolution
  - POST request payload example `[{ exporterName: 'Acme' }]`
  - Before site creation we check if site already exists
  - If siteService.getSiteStatusByExporterName doesn't throw SiteNotFoundException then response is returned to client
    - Most likely it is information about status of existing Site.
  - If we get SiteNotFoundException, then new site creation can start.
    - Generates new site id using MDM API.
    - Service does MS graph api POST call, with new SiteId for field URL and ExporterName for Title and other fields.
    - Response is based on sharepoint field Sitestatus. In most cases it's value will be `Provisioning` and http status code will be `202`


src/helpers/ukef-id.type.ts
```diff
@@ -0,0 +1,5 @@
+// Defining multiple String Literal Types, this doesn't support constants.
+type SiteUkefIdStart = '070';
```
AlexBramhill: It's a bit surprising that this is `070` and not `0070`. Is this correct?

src/modules/site/dto/create-site-request.dto.ts
```diff
@@ -0,0 +1,19 @@
...
+  @ValidatedStringApiProperty({
+    description: 'The name of the exporter used in the deal.',
+    example: 'Example Name Limited',
+    minLength: 1,
+    maxLength: 250,
+    pattern: EXPORTER_NAME.REGEX,
+  })
+  readonly exporterName: string;
+
+  constructor(exporterName: string) {
```
AlexBramhill: Do we need/are we using this constructor?

src/modules/site/exception/site-id-creation-failed.exception.ts
```diff
@@ -0,0 +1,8 @@
+import { SiteException } from './site.exception';
+
+export class SiteIdCreationFailedException extends SiteException {
```
AlexBramhill: Could we have the standard tests for this please?

src/modules/site/site.controller.test.ts
```diff
@@ -87,4 +96,98 @@ describe('SiteController', () => {
       await expect(responsePromise).rejects.toThrow(error);
     });
   });
+
+  describe('createSite', () => {
+    const { siteStatusByExporterNameResponse } = new getSiteStatusByExporterNameGenerator(valueGenerator).generate({ numberToGenerate: 1 });
```
AlexBramhill: Feels a bit surprising to rely on this generator here -- it'd be good to abstract out the `suteStatusByExporterNameResponse` and call this new generator in both `getSiteStatusByExporterNameGenerator` and `CreateSiteGenerator`. What do you think?

src/modules/site/site.controller.test.ts
```diff
@@ -87,4 +96,98 @@ describe('SiteController', () => {
...
+      },
+      {
+        status: ENUMS.SITE_STATUSES.PROVISIONING,
+        expectedStatusCode: HttpStatusCode.Accepted,
+      },
+      {
+        status: ENUMS.SITE_STATUSES.CREATED,
+        expectedStatusCode: HttpStatusCode.Ok,
+      },
+    ])('returns a status code of $expectedStatusCode and the expected response if site status is "$status"', async ({ status, expectedStatusCode }) => {
```
AlexBramhill: ('returns a status code of $expectedStatusCode and the expected response if **the created** site status is "$status"')

maybe? Then we can remove the comment as we've made it clear that this test does belong here

src/modules/site/site.controller.test.ts
```diff
@@ -87,4 +96,98 @@ describe('SiteController', () => {
...
+
+      await siteController.createSite(createSiteRequest, responseMock);
+
+      expect(responseMock.json).toHaveBeenCalledTimes(1);
+      expect(responseMock.json).toHaveBeenCalledWith(modifiedSiteStatusByExporterNameResponse);
+      expect(responseMock.status).toHaveBeenCalledTimes(1);
+      expect(responseMock.status).toHaveBeenCalledWith(expectedStatusCode);
+    });
+
+    it('calls create site and returns a status code 202 if site service throws a SiteNotFoundException', async () => {
```
AlexBramhill: Is this correct?

AlexBramhill: Ah I think I get this now -- I feel this test could have a rename.

`calls create site and returns a status code of 202 if the site service request to get site status throws a SiteNotFoundException`

src/modules/site/site.controller.test.ts
```diff
@@ -87,4 +96,98 @@ describe('SiteController', () => {
...
+      expect(responseMock.json).toHaveBeenCalledTimes(1);
+      expect(responseMock.json).toHaveBeenCalledWith({
+        siteId: siteId,
+        status: ENUMS.SITE_STATUSES.PROVISIONING,
+      });
+      expect(responseMock.status).toHaveBeenCalledTimes(1);
+      expect(responseMock.status).toHaveBeenCalledWith(HttpStatusCode.Accepted);
+    });
+
+    it('throws an error if site service throws an error which is not SiteNotFoundException', async () => {
```
AlexBramhill: `throws an error if the site service request to get site status throws an error which is not SiteNotFoundException` maybe?

src/modules/site/site.controller.test.ts
```diff
@@ -87,4 +96,98 @@ describe('SiteController', () => {
...
+      when(siteGetSiteStatusByExporterName).calledWith(exporterName).mockRejectedValueOnce(siteNotFoundError);
+      when(siteCreateSiteId).calledWith().mockReturnValue(siteId);
+      when(siteCreateSite).calledWith(exporterName, siteId).mockResolvedValueOnce(createSiteResponse[0]);
+
+      await siteController.createSite(createSiteRequest, responseMock);
+
+      expect(siteCreateSite).toHaveBeenCalledTimes(1);
+      expect(responseMock.json).toHaveBeenCalledTimes(1);
+      expect(responseMock.json).toHaveBeenCalledWith({
+        siteId: siteId,
```
AlexBramhill: I think this can just be `siteId`?

src/modules/site/site.controller.test.ts
```diff
@@ -87,4 +96,98 @@ describe('SiteController', () => {
...
+
+      const responsePromise = siteController.createSite(createSiteRequest, responseMock);
+
+      await expect(responsePromise).rejects.toThrow(error);
+    });
+
+    it('throws error SiteIdCreationFailedException if site service throws MdmException', async () => {
+      const siteNotFoundError = new SiteNotFoundException(`Site not found for exporter name: ${exporterName}`);
+      const mdmError = new MdmException('Error message');
+      const expectedFinalError = new SiteIdCreationFailedException('Failed to create a site ID');
```
AlexBramhill: `expectedResponseError`?

src/modules/site/site.controller.test.ts
```diff
@@ -1,10 +1,13 @@
 import { ENUMS } from '@ukef/constants';
+import { CreateSiteGenerator } from '@ukef-test/support/generator/create-site-generator';
```
AlexBramhill: I'd be tempted to split these two controller tests into two files for ease of management

src/modules/site/site.controller.ts
```diff
@@ -37,6 +43,47 @@ export class SiteController {
   @ApiBadRequestResponse({ description: 'Bad request.' })
   async getSiteStatusByExporterName(@Query() query: GetSiteStatusByExporterNameQueryDto, @Res() res: Response): Promise<void> {
     const getSiteStatusByExporterNameResponse = await this.service.getSiteStatusByExporterName(query.exporterName);
+    this.setSiteStatusResponse(res, getSiteStatusByExporterNameResponse);
```
AlexBramhill, line 43: nice!

src/modules/site/site.controller.ts
```diff
@@ -37,6 +43,47 @@ export class SiteController {
...
+      this.setSiteStatusResponse(res, existingSite);
+    } catch (error) {
+      if (error instanceof SiteNotFoundException) {
+        // Site doesn't exists, add item to sharepoint create site list.
+        try {
+          const siteId = await this.service.createSiteId();
+          const createSiteResponse = await this.service.createSite(exporterName, siteId);
+          this.setSiteStatusResponse(res, createSiteResponse);
+        } catch (error) {
+          if (error instanceof MdmException) {
```
AlexBramhill: Could we use an interceptor for this like we have done for the `get` endpoint?

src/modules/site/site.controller.ts
```diff
@@ -37,6 +43,47 @@ export class SiteController {
...
+    type: CreateSiteResponse,
+    isArray: false,
+  })
+  @ApiUnauthorizedResponse({ description: 'Unauthorized' })
+  @ApiInternalServerErrorResponse({ description: 'An internal server error has occurred.' })
+  @ApiBadRequestResponse({ description: 'Bad request.' })
+  async createSite(@ValidatedArrayBody({ items: CreateSiteRequestItem }) createSiteDto: CreateSiteRequest, @Res() res: Response): Promise<void> {
+    const exporterName = createSiteDto[0].exporterName;
+    try {
+      const existingSite = await this.service.getSiteStatusByExporterName(exporterName);
```
AlexBramhill: @lmarrai-sw might have an opinion on this -- I'd be tempted move this logic to `siteService` as a `createSiteIfDoesNotExist` and simplify the logic of this controller down. There's a lot of nesting here, which I think that moving to siteService will clear up, as well as changing to interceptors

src/app.ts
```diff
@@ -32,9 +32,6 @@ export class App {
         whitelist: true,
         transform: true,
         forbidNonWhitelisted: true,
```
lmarrai-sw: I know it's not part of your change @avaitonis, but we removed `forbidNonWhitelisted: true,` in TFS

test/support/mocks/mdm.service.mock.ts
```diff
@@ -0,0 +1,22 @@
...
+    this.createNumbers = jest.fn();
+  }
+
+  mockSuccessfulReturnValue(siteId: UkefSiteId) {
+    when(this.createNumbers)
+      .calledWith(expect.anything())
+      .mockReturnValueOnce([{ maskedId: siteId }]);
+  }
+
+  mockMdMCallError() {
```
lmarrai-sw: Very small typo, but did we mean `Mdm` instead of `MdM` here?

src/modules/site/site.controller.ts
avaitonis: For client we return 500 `Internal server error` without any details.
@luke you suggested rewrap MdmException `Failed to create numbers in MDM` to SiteIdCreationFailedException `Failed to create a site ID`. Is this required at all? 
It's a bit messy in controller, moving to existing SiteExceptionTransformInterceptor is easy, but testing interceptor is not straight forward.

AlexBramhill: I believe the usual way we look to test interceptors would be at the api-test level -- would that make it easier?

test/support/mocks/mdm.service.mock.ts
```diff
@@ -0,0 +1,22 @@
+import { UkefSiteId } from '@ukef/helpers/ukef-id.type';
```
AlexBramhill, line 1: I'd be tempted as mdm is using simple http endpoints to use `nock` here ala TFS. This allows us the check the request is being sent as expected, and is setting the headers etc as needed.

The reason we don't do this for Graph service is it uses an external package to make these calls. In that case, we don't care about _what_ calls it is making, rather that we are calling the external package correctly, and we mock the response based on that.

src/modules/site/site.controller.ts
```diff
@@ -37,6 +43,47 @@ export class SiteController {
...
+  @ApiResponse({
+    status: HttpStatusCode.FailedDependency,
+    description: 'Site has failed to be created',
+    type: CreateSiteResponse,
+    isArray: false,
+  })
+  @ApiUnauthorizedResponse({ description: 'Unauthorized' })
+  @ApiInternalServerErrorResponse({ description: 'An internal server error has occurred.' })
+  @ApiBadRequestResponse({ description: 'Bad request.' })
+  async createSite(@ValidatedArrayBody({ items: CreateSiteRequestItem }) createSiteDto: CreateSiteRequest, @Res() res: Response): Promise<void> {
```
lmarrai-sw, line 59: Should the return type be `Promise<CreateSiteResponse>` instead?

test/common-tests/request-field-validation-api-tests/string-field-validation-api-tests.ts
```diff
@@ -0,0 +1,257 @@
...
+            error: 'Bad Request',
+            message: expect.arrayContaining([`${fieldName} must be one of the following values: ${Object.values(theEnum).join(', ')}`]),
+            statusCode: 400,
+          });
+        });
+      } else {
+        it(`returns a 400 response if ${fieldName} is not present`, async () => {
+          const { [fieldNameSymbol]: _removed, ...requestWithoutTheField } = requestBodyItem;
+
+          const { status, body } = await makeRequest([requestWithoutTheField]);
```
lmarrai-sw: I noticed a small mistake in TFS this morning: this should be `makeRequest(prepareModifiedRequest(requestWithoutTheField))` instead

Can you fix it here please @avaitonis and I'll fix it in TFS?

lmarrai-sw (commented):
Thanks for this @avaitonis

src/helpers/ukef-id.type.ts
avaitonis: Thank you, changed to `0070`.

src/modules/site/dto/create-site-request.dto.ts
avaitonis: Removed constructor

src/modules/site/exception/site-id-creation-failed.exception.ts
avaitonis: We had chat with Luke and decided to remove SiteIdCreationFailedException,
Would site id creation fail, then logs will have MdmException.

src/modules/site/site.controller.test.ts
avaitonis: Post endpoint is using get site service, so get site generator is also used quite a lot. It's possible to abstract generators, but I am leaving them separate for this PR.

src/modules/site/site.controller.test.ts
avaitonis: Thank you, changed test description.

src/modules/site/site.controller.test.ts
avaitonis: Thank you, changed test description.

src/modules/site/site.controller.test.ts
avaitonis: Thank you, changed test description.

src/modules/site/site.controller.test.ts
avaitonis: Changed to shorter version.

src/modules/site/site.controller.test.ts
avaitonis: Thanks, changed

src/modules/site/site.controller.test.ts
avaitonis: Splitted.

src/modules/site/site.controller.ts
avaitonis: Thanks Alex, it was quite messy. 

Interceptor worked fine, but was a bit tricky to test. At api-test both exceptions returned same generic 500 error. It was a way to test it, but not so straight forward for me.

We decided to remove SiteIdCreationFailedException.

src/modules/site/site.controller.ts
avaitonis: Thanks, logic is moved to `siteService`.

test/support/mocks/mdm.service.mock.ts
avaitonis: Sounds like good practice, but mdm call is quite simple, so not sure how important it is.

Created new user story APIM-478 so we can prioritise.

src/app.ts
avaitonis: OK, I removed forbidNonWhitelisted. One get site test was testing that unknown query parameters not allowed, this test had to be changed.

We could have forbidNonWhitelisted: true, looks like nice extra security measure.

test/support/mocks/mdm.service.mock.ts
avaitonis: Changed

src/modules/site/site.controller.ts
avaitonis: To set http status we need to change object res. Same object is used for response body. It is same approach as in GET site endpoint.

So this function directly is not returning anything.

test/common-tests/request-field-validation-api-tests/string-field-validation-api-tests.ts
avaitonis: Fixed

src/modules/site/site.controller.create-site.test.ts
```diff
@@ -0,0 +1,84 @@
...
+        status: ENUMS.SITE_STATUSES.PROVISIONING,
+        expectedStatusCode: HttpStatusCode.Accepted,
+      },
+      {
+        status: ENUMS.SITE_STATUSES.CREATED,
+        expectedStatusCode: HttpStatusCode.Ok,
+      },
+    ])(
+      'returns a status code of $expectedStatusCode and the expected response if the created site status is "$status"',
+      async ({ status, expectedStatusCode }) => {
```
AlexBramhill, line 51: Nice use of each by the way :)

src/modules/site/site.controller.ts
```diff
@@ -37,6 +40,28 @@ export class SiteController {
...
+    const siteStatus = await this.service.createSiteIfDoesNotExist(createSiteDto[0].exporterName);
```
AlexBramhill, line 60: Great job refactoring this

test/site/create-post.api-test.ts
```diff
@@ -0,0 +1,251 @@
+import { ENUMS, EXPORTER_NAME } from '@ukef/constants';
```
AlexBramhill: Could the name of this file follow convention from tfs please? (feel free to resolve this comment once updated)

test/site/create-post.api-test.ts
```diff
@@ -0,0 +1,251 @@
...
+    givenRequestWouldOtherwiseSucceed: () => {
+      mockGraphClientService.mockSuccessfulGraphApiCallWithPath(graphServiceGetParams.path);
+    },
+    givenGraphServiceCallWillThrowError: (error: Error) => {
+      mockGraphClientService.mockUnsuccessfulGraphGetCall(error);
+    },
+    makeRequest: () => api.post(`/api/v1/sites`, createSiteRequest),
+  });
+
+  const statusCodeTestInputs = [
```
AlexBramhill: Not needed for now -- but if this test files gets unmanageably large, I think we could probably pull this out into a test part, as it's used in the `get-site-status` endpoint tests too

test/site/get-site-status-by-exporter-name.api-test.ts
```diff
@@ -105,7 +105,7 @@ describe('getSiteStatusByExporterName', () => {
     expect(body).toStrictEqual({ message: 'Not found', statusCode: 404 });
   });
 
-  it('returns a 400 with message containing "property does not exist" if query property does not exist', async () => {
+  it("returns a 400 because query parameter is missing, but incorrect query parameter doesn't trigger error", async () => {
```
AlexBramhill: Any reason we are repurposing this test? I'd expect the old test still to pass?

AlexBramhill (changes requested):
Looking nice! Only a couple of comments. (I'd wait until another approval to merge)

test/site/get-site-status-by-exporter-name.api-test.ts
avaitonis: Luke highlighted that ValidationPipe configuration in TFS is different, so I removed forbidNonWhitelisted: true. This change broke this GET site test.

test/site/create-post.api-test.ts
avaitonis: Changed to post-site.api-test.ts . Old name didn't make sense, I just didn't notice.

src/modules/site/site.controller.create-site.test.ts
avaitonis: All yours :)

test/site/get-site-status-by-exporter-name.api-test.ts
lmarrai-sw: I think we should either just delete this test or make it "if you use the _correct_ query parameter _and_ an unexpected one as well then it still succeeds/you get a 200 response" - otherwise it's a bit confusing what exactly this test is for :)

src/app.ts
lmarrai-sw: `whitelist: true` is a good security measure here and `forbidNonWhitelisted: false` (or missing) is good for client flexibility and release flexibility (@avaitonis and I just discussed this on a call)

src/modules/site/site.controller.test.ts
lmarrai-sw: I'm happy with this

lmarrai-sw (dismissed):
LGTM, thanks @avaitonis

test/site/get-site-status-by-exporter-name.api-test.ts
avaitonis: Changed test to "if you use the correct query parameter and an unexpected one as well, then it still succeeds/you get a 202 response". 202 "Provisioning" is default status in get site generator.

lmarrai-sw (dismissed):
LGTM, thanks @avaitonis

test/site/get-site-status-by-exporter-name.api-test.ts
```diff
@@ -92,36 +92,41 @@ describe('getSiteStatusByExporterName', () => {
     expect(body).toStrictEqual(modifiedSiteStatusByExporterNameResponse);
   });
 
-  it('returns 404 with an empty object response if the site does not exist in sharepoint', async () => {
+  it('if you use the correct query parameter and an unexpected one as well, then it still succeeds/you get a 202 response', async () => {
+    const {
```
AlexBramhill: Is there any reason why this is using this constructor here? We can probably just reference the main generator by the looks of it?

AlexBramhill (dismissed):
lgtm (one slight question)!

test/site/get-site-status-by-exporter-name.api-test.ts
avaitonis: Good point, removed redundant generator call.

AlexBramhill (approved):
Great work!

PR #69, opened 2023-06-27
feat(APIM-450): add temporary endpoint to check file exists and does not exceed max size

oscar-richardson-softwire:
Already reviewed in https://github.com/UK-Export-Finance/estore-api/pull/46


lmarrai-sw (approved):
Thanks @oscar-richardson-softwire!

PR #70, opened 2023-06-27
feat(APIM-136): add custodian module for creating and provisioning items

lmarrai-sw:
## Introduction
As part of APIM-136 and other similar tickets, we need to send a POST request to Custodian to create folders in Sharepoint. I wanted to get that merged before the rest of APIM-136 so that the similar tickets can re-use it easily.

## Resolution
- Added a `custodian` module and config for sending the POST request to Custodian
- Refactored the `MdmService` to use a new `HttpClient` so that I didn't need to duplicate its code for the new `CustodianService`

## Misc
- I created repeatable tests for config parsing that `mdm.config.test.ts` and `custodian.config.test.ts` can use so we don't need to duplicate that code


src/modules/custodian/custodian.service.ts
```diff
@@ -0,0 +1,29 @@
...
+import { CustodianException } from './exception/custodian.exception';
+
+@Injectable()
+export class CustodianService {
+  private readonly httpClient: HttpClient;
+  constructor(httpService: HttpService) {
+    this.httpClient = new HttpClient(httpService);
+  }
+
+  // TODO APIM-136: Change this to return Promise<void> if we do not end up using the response data
```
lmarrai-sw, line 17: I will decide on this TODO later in the ticket

src/modules/custodian/dto/custodian-create-and-provision-request.dto.ts
```diff
@@ -0,0 +1,22 @@
...
+  TemplateId: string;
+  ParentId?: number;
+  InterestedParties?: string;
+  Secure?: boolean;
+  DoNotSubscribeInterestedParties?: boolean;
+  Links?: []; // Custodian supports sending an array of links, but we do not use this functionality.
+  FormButton?: string;
+  HasAttachments?: boolean;
+  Metadata?: MetadataItem[];
+  ConnectedListMetadata?: []; // TODO APIM-136: Mulesoft sends an empty array for this field but it is not mentioned in Custodian's documentation.
```
lmarrai-sw: @avaitonis I wanted to double-check with you, but I assume you don't have any reason why we should keep this field if it's not documented by Custodian? (https://custodian.repstor.com/swagger/ui/index)

.env.sample
```diff
@@ -32,18 +32,16 @@ APIM_MDM_VALUE=
...
 DTFS_STORAGE_FILE_SHARE_NAME=
 DTFS_STORAGE_EXAMPLE_FILE_NAME=
 DTFS_STORAGE_EXAMPLE_FILE_LOCATION_PATH=
+
+# CUSTODIAN
+CUSTODIAN_BASE_URL=
+CUSTODIAN_API_KEY_HEADER_NAME=
+CUSTODIAN_API_KEY_HEADER_VALUE=
+CUSTODIAN_MAX_REDIRECTS=
+CUSTODIAN_TIMEOUT=
```
lmarrai-sw: @abhi-markan when you get a chance, can you please set up these variable in GitHub? I can tell you what the values need to be

src/modules/custodian/dto/custodian-create-and-provision-request.dto.ts
avaitonis: We can try to remove this, maybe it is legacy field.

avaitonis (dismissed):
LGTM

.env.sample
lmarrai-sw: I'm going to raise a Jira ticket for Abhi to do this if needed, following our new process

src/modules/custodian/dto/custodian-create-and-provision-request.dto.ts
lmarrai-sw: Thanks, I've removed it

avaitonis (approved):
LGTM

PR #71, opened 2023-06-27
fix(iac): added CRLF escape

abhi-markan:
## Introduction
IaC new line was no escaped

## Resolution
* Added `\` to escape CRLF

## Miscellaneous
* Dependencies update


oscar-richardson-softwire (approved):
LGTM

PR #91, opened 2023-07-05
fix(apim 480): post terms facilities error handling

francescastocco:
## Introduction
When the POST /terms/facilities endpoint is called and the item already exists we receive a 500 internal server error, but want to receive a 200 instead.

## Resolution
- Created a known-error `TermsFacilityExistsException` to catch the error
- When we have an instance of this known error we return a 200 with the message `Facility Term already exists`
- Updated search for known/shared errors to use error code or message

## Misc
I checked for other potential known-errors, eg. Invalid characters or character limit, but we have validation on the request body such that it wouldn't accept these anyway


src/modules/terms/terms.controller.test.ts
```diff
@@ -44,32 +46,58 @@ describe('TermsController', () => {
...
-
-      when(mockService).calledWith(term.id).mockResolvedValue(result);
-
-      const response = await controller.postFacilityToTermStore([term]);
+    it.each([
+      {
+        expectedResult: { message: ENUMS.CREATE_TERM_FOR_FACILITY_RESPONSES.FACILITY_TERM_CREATED },
+        expectedStatusCode: HttpStatusCode.Created,
+      },
+
```
lmarrai-sw: Is the blank line here deliberate?

lmarrai-sw (dismissed):
LGTM thanks @francescastocco

IgnatG (dismissed):

lmarrai-sw (approved):

PR #92, opened 2023-07-05
feat(APIM-138): add endpoint to upload a file in the deal folder to Sharepoint

oscar-richardson-softwire:
Draft PR for functionality (no tests) for [APIM-138](https://ukef-dtfs.atlassian.net/browse/APIM-138).


.env.sample
```diff
@@ -45,3 +50,18 @@ CUSTODIAN_API_KEY_HEADER_NAME=
...
 CUSTODIAN_MAX_REDIRECTS=
 CUSTODIAN_TIMEOUT=
+
+
+# MDM
+MDM_BASE_URL=
+MDM_API_KEY_HEADER_NAME=
+MDM_API_KEY_HEADER_VALUE=
+MDM_MAX_REDIRECTS=
+MDM_TIMEOUT=
```
lmarrai-sw: I think these are duplicates of the `APIM MDM` ones

.env.sample
```diff
@@ -45,3 +50,18 @@ CUSTODIAN_API_KEY_HEADER_NAME=
...
+
+
+# STORAGE
+DTFS_STORAGE_ACCOUNT_NAME=
+DTFS_STORAGE_ACCOUNT_KEY=
+DTFS_STORAGE_EXAMPLE_FILE_NAME=
+DTFS_STORAGE_EXAMPLE_FILE_LOCATION_PATH=
```
lmarrai-sw: I think these already exist in the file

src/modules/deal-folder/deal-folder.module.ts
```diff
@@ -0,0 +1,14 @@
...
+import { DtfsStorageModule } from '../dtfs-storage/dtfs-storage.module';
+import { GraphModule } from '../graph/graph.module';
+import { DealFolderController } from './deal-folder.controller';
+import { DealFolderService } from './deal-folder.service';
+
+@Module({
+  imports: [DtfsStorageModule, GraphModule],
+  controllers: [DealFolderController],
+  providers: [DealFolderService],
+  exports: [],
```
lmarrai-sw: You don't need to list `exports` if it's an empty array

src/modules/dtfs-storage-client/dtfs-storage-client.module.ts
```diff
@@ -0,0 +1,10 @@
+import { Module } from '@nestjs/common';
+
+import DtfsStorageClientService from './dtfs-storage-client.service';
+
+@Module({
+  imports: [],
```
lmarrai-sw: You can remove `imports` here

src/modules/deal-folder/dto/upload-file-in-deal-folder-response.dto.ts
```diff
@@ -0,0 +1,10 @@
+import { EXAMPLES } from '@ukef/constants/examples.constant';
+import { ValidatedStringApiProperty } from '@ukef/decorators/validated-string-api-property.decorator';
+
+export class UploadFileInDealFolderResponseDto {
+  @ValidatedStringApiProperty({
+    description: '', // TODO (APIM-138): add description.
```
lmarrai-sw: Highlighting the TODO :)

src/modules/deal-folder/deal-folder.controller.ts
```diff
@@ -0,0 +1,34 @@
...
+export class DealFolderController {
+  constructor(private readonly service: DealFolderService) {}
+
+  @Post('sites/:siteId/deals/:dealId/documents')
+  @ApiOperation({ summary: 'Upload a file in the deal folder.' })
+  @ApiBody({
+    type: UploadFileInDealFolderRequestItem,
+    isArray: true,
+  })
+  @ApiCreatedResponse({ description: 'File has been uploaded.', type: UploadFileInDealFolderResponseDto, isArray: false })
```
lmarrai-sw: You can skip `isArray: false` here, that's the default

src/modules/dtfs-storage/dtfs-storage-file.service.ts
```diff
@@ -1,39 +1,41 @@
...
-    @Inject(DtfsConfig.KEY)
-    { baseUrl, accountName, accountKey }: ConfigType<typeof DtfsConfig>,
-  ) {
-    this.baseUrl = baseUrl;
-    this.storageSharedKeyCredential = new StorageSharedKeyCredential(accountName, accountKey);
+  private readonly dtfsStorageClientService: DtfsStorageClientService;
+
+  constructor(dtfsStorageClientService: DtfsStorageClientService) {
+    this.dtfsStorageClientService = dtfsStorageClientService;
   }
```
lmarrai-sw: You can do this with
```
constructor(private readonly dtfsStorageClientService: DtfsStorageClientService) { }

File: src/constants/document-type-mapping.constant.ts [hunk 8]
```diff
@@ -0,0 +1,32 @@
+import { ENUMS } from './enum.constant';
+
+const estoreDocumentTypeIdApplication = process.env.SHAREPOINT_ESTORE_DOCUMENT_TYPE_ID_APPLICATION;
+const estoreDocumentTypeIdFinancialStatement = process.env.SHAREPOINT_ESTORE_DOCUMENT_TYPE_ID_FINANCIAL_STATEMENT;
+const estoreDocumentTypeIdBusinessInformation = process.env.SHAREPOINT_ESTORE_DOCUMENT_TYPE_ID_BUSINESS_INFORMATION;
```
lmarrai-sw: Loading env vars directly like this doesn't feel right to me, and doesn't fit with the rest of the project.

Maybe we load the env vars into a config object and inject the config object into a "mapper" class that takes care of the mapping below? So code that needs the mapping would ask the mapper class what the correct title and type id is for a given document type.

Let me know if you want to talk about what I mean

src/modules/deal-folder/deal-folder.service.ts
```diff
@@ -0,0 +1,175 @@
...
+      .then((response) => response.value)
+      .then((resources) => resources.find((resource) => resource.name === resourceName)) // This is necessary because the 'filter' query parameter does not currently support the 'name' field on lists/list items (see comment from Microsoft employee https://learn.microsoft.com/en-us/answers/questions/980144/graph-api-sharepoint-list-filter-by-createddatetim).
+      .then((resource) => resource.id); // Additionally, the 'filter' query parameter is not supported at all by the 'List available drives' endpoint (see https://learn.microsoft.com/en-us/graph/api/drive-list?view=graph-rest-1.0&tabs=http#optional-query-parameters for a list of supported parameters).
+  }
+
+  private constructRequestBodyToUpdateFileInfo(documentType: string): {
+    Title: string;
+    Document_x0020_Status: string;
+  } {
+    const { documentTitle, documentTypeId } = DOCUMENT_TYPE_MAPPING[documentType];
```
lmarrai-sw: We've got warning here
> Variable Assigned to Object Injection Sink
  
I think introducing a mapping class as suggested above might fix that too

src/modules/deal-folder/deal-folder.service.ts
```diff
@@ -0,0 +1,175 @@
...
+
+    return {
+      fileUpload: fileLeafRef,
+    };
+  }
+
+  private async checkFileIsNotTooLargeAndReturnSizeInBytes(fileName: string, fileLocationPath: string): Promise<number> {
+    const fileSizeInBytes = await this.dtfsStorageFileService.getFileProperties(fileName, fileLocationPath).then((response) => {
+      return response.contentLength;
+    });
```
lmarrai-sw: Shall we use `(response) => response.contentLength` instead? (removing the curly braces in the anonymous function)

src/modules/deal-folder/deal-folder.service.ts
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  private getItemIdAndFileLeafRefByWebUrlRegExp(ukefSiteId: string, listId: string, webUrlRegExp: RegExp): Promise<{ itemId: string; fileLeafRef: string }> {
+    return this.graphService
+      .get<{ value: { webUrl: string; id: string; fields: { FileLeafRef: string } }[] }>({
+        path: `sites/${this.config.ukefSharepointName}:/sites/${ukefSiteId}:/lists/${listId}/items`,
+        expand: 'fields',
+      })
+      .then((response) => response.value)
+      .then((list) => list.find((item) => webUrlRegExp.test(item.webUrl))) // This is necessary because the 'filter' query parameter does not currently support the 'webUrl' field on lists/list items (see comment from Microsoft employee https://learn.microsoft.com/en-us/answers/questions/980144/graph-api-sharepoint-list-filter-by-createddatetim).
```
lmarrai-sw: Is this comment still 100% accurate? Isn't it also because we need to use regex, or does `$filter` support regex too?

src/modules/deal-folder/deal-folder.service.ts
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  private async constructUrlToCreateUploadSession(fileName: string, dealId: string, buyerName: string, ukefSiteId: string): Promise<string> {
+    const sharepointSiteId = await this.getSharepointSiteIdByUkefSiteId(ukefSiteId);
+    const driveId = await this.getResourceIdByName(ukefSiteId, CASE_LIBRARY.DRIVE_NAME, ENUMS.SHAREPOINT_RESOURCE_TYPES.DRIVE);
+    // Getting the sharepointSiteId () and driveId appears to necessary because using the ukefSiteId and drive name only appears to work when accessing drives,
+    // not items *within* drives.
+    const fileDestinationPath = `${buyerName}/D${dealId}`;
+
+    return `https://graph.microsoft.com/v1.0/sites/${sharepointSiteId}/drives/${driveId}/root:/${fileDestinationPath}/${fileName}:/createUploadSession`;
```
lmarrai-sw: We should probably move https://graph.microsoft.com/v1.0` into an env variable instead, so that we can point to different versions of the API (and mock APIs) easily

src/modules/deal-folder/deal-folder.service.ts
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  private getResourceIdByName(ukefSiteId: string, resourceName: string, sharepointResourceType: SharepointResourceTypeEnum): Promise<string> {
+    return this.graphService
+      .get<{ value: { name: string; id: string }[] }>({
+        path: `sites/${this.config.ukefSharepointName}:/sites/${ukefSiteId}:/${sharepointResourceType}s`,
+      })
+      .then((response) => response.value)
+      .then((resources) => resources.find((resource) => resource.name === resourceName)) // This is necessary because the 'filter' query parameter does not currently support the 'name' field on lists/list items (see comment from Microsoft employee https://learn.microsoft.com/en-us/answers/questions/980144/graph-api-sharepoint-list-filter-by-createddatetim).
+      .then((resource) => resource.id); // Additionally, the 'filter' query parameter is not supported at all by the 'List available drives' endpoint (see https://learn.microsoft.com/en-us/graph/api/drive-list?view=graph-rest-1.0&tabs=http#optional-query-parameters for a list of supported parameters).
```
lmarrai-sw: As there are quite long comments, it might be better to have them on their own lines instead of at the end of a line

src/modules/deal-folder/deal-folder.service.ts
```diff
@@ -0,0 +1,175 @@
...
+      requestBody: requestBodyToUpdateFileInfo,
+    });
+
+    return fileLeafRef;
+  }
+
+  private async constructUrlToCreateUploadSession(fileName: string, dealId: string, buyerName: string, ukefSiteId: string): Promise<string> {
+    const sharepointSiteId = await this.getSharepointSiteIdByUkefSiteId(ukefSiteId);
+    const driveId = await this.getResourceIdByName(ukefSiteId, CASE_LIBRARY.DRIVE_NAME, ENUMS.SHAREPOINT_RESOURCE_TYPES.DRIVE);
+    // Getting the sharepointSiteId () and driveId appears to necessary because using the ukefSiteId and drive name only appears to work when accessing drives,
```
lmarrai-sw: Is `()` deliberate in this comment?

src/modules/deal-folder/deal-folder.service.ts
```diff
@@ -0,0 +1,175 @@
...
+    // not items *within* drives.
+    const fileDestinationPath = `${buyerName}/D${dealId}`;
```
lmarrai-sw: I thought the deal folder name had a space between D and the deal id - is this something different?

.env.sample
oscar-richardson-softwire: Removed

.env.sample
oscar-richardson-softwire: Changed

src/constants/examples.constant.ts
```diff
@@ -9,4 +9,7 @@ export const EXAMPLES = {
   FILE_SIZE: MAX_FILE_SIZE_BYTES,
   FILE_NAME: process.env.DTFS_STORAGE_EXAMPLE_FILE_NAME,
   FILE_LOCATION_PATH: process.env.DTFS_STORAGE_EXAMPLE_FILE_LOCATION_PATH,
+  BUYER_NAME: 'Shanghai Ship and Shipping Research Institute T',
```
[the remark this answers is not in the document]
oscar-richardson-softwire: I copied this example from Mulesoft, but I don't see any disadvantage of just using `Example Buyer`, so I've changed it to that

src/modules/deal-folder/deal-folder.module.ts
oscar-richardson-softwire: Removed

src/modules/dtfs-storage-client/dtfs-storage-client.module.ts
oscar-richardson-softwire: Removed

src/modules/deal-folder/deal-folder.controller.ts
oscar-richardson-softwire: Removed

src/modules/dtfs-storage/dtfs-storage-file.service.ts
oscar-richardson-softwire: Nice, thanks, changed

src/modules/deal-folder/deal-folder.service.ts
oscar-richardson-softwire: Good spot, changed

src/constants/examples.constant.ts
```diff
@@ -9,4 +9,7 @@ export const EXAMPLES = {
...
+  DOCUMENT_TYPE: ENUMS.DOCUMENT_TYPES.AUDITED_FINANCIAL_STATEMENTS,
+  FILE_UPLOAD: '', // TODO (APIM-138): add example.
```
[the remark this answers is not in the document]
oscar-richardson-softwire: Actioned and removed

src/modules/deal-folder/dto/upload-file-in-deal-folder-response.dto.ts
oscar-richardson-softwire: Actioned and removed

src/constants/document-type-mapping.constant.ts
oscar-richardson-softwire: Good suggestion, I've implemented this

src/modules/deal-folder/deal-folder.service.ts
oscar-richardson-softwire: I used a switch statement in the method on the mapping class which fixed this

src/modules/deal-folder/deal-folder.service.ts
oscar-richardson-softwire: Done

src/modules/deal-folder/deal-folder.service.ts
oscar-richardson-softwire: Changed to multi-line comments on their own lines

src/modules/deal-folder/deal-folder.service.ts
oscar-richardson-softwire: Good spot, I meant to put the format of the sharepointSiteId in those brackets, which I've added now

src/modules/deal-folder/deal-folder.service.ts
oscar-richardson-softwire: Great catch! I missed this space when reading the Mulesoft code. Changed now

src/modules/deal-folder/deal-folder.service.ts
oscar-richardson-softwire: I've now removed the regex as it is no longer needed after the 'await' fix yesterday, so this comment is 100% accurate.

src/modules/deal-folder/deal-folder.service.ts
```diff
@@ -0,0 +1,182 @@
...
+    const fileLeafRef = await this.updateFileInformationInSharepoint(fileName, dealId, buyerName, ukefSiteId, documentType);
+
+    return {
+      fileUpload: fileLeafRef,
+    };
+  }
+
+  private async checkFileIsNotTooLargeAndReturnSizeInBytes(fileName: string, fileLocationPath: string): Promise<number> {
+    const fileSizeInBytes = await this.dtfsStorageFileService.getFileProperties(fileName, fileLocationPath).then((response) => response.contentLength);
+    if (fileSizeInBytes > MAX_FILE_SIZE_BYTES) {
```
lmarrai-sw: I think `MAX_FILE_SIZE_BYTES` has been renamed to `DTFS_MAX_FILE_SIZE_BYTES` on main, so watch out for that when you rebase

lmarrai-sw (dismissed):

src/modules/graph/graph.service.ts
```diff
@@ -59,13 +68,45 @@ export class GraphService {
...
+      createGraphError({
+        error,
+        messageForUnknownError: 'An unexpected error occurred.',
+        knownErrors: [],
+      });
+    }
+  }
+
+  async uploadFile(file: NodeJS.ReadableStream, fileSizeInBytes: number, fileName: string, urlToCreateUploadSession: string): Promise<UploadResult> {
+    const uploadSession: LargeFileUploadSession = await LargeFileUploadTask.createUploadSession(this.client, urlToCreateUploadSession, {});
```
oscar-richardson-softwire: @avaitonis The object that we pass as the last argument allows us to set options, notably the behaviour when a file with the same name already exists in the location. What do we want this to be? The options are overwrite the existing file, rename the new file `conflictingFileName N.extension` when it is the N+1th file with that name, or fail (which returns a `409` error - would we want to map this to a `400`?). 

In the Mulesoft code, it appears that in the SharePoint connector `file-add` operation the `Overwrite` boolean isn't set, and the default is false, but the docs don't make it clear what happens in this case (i.e., does it fail or rename the new file?) https://docs.mulesoft.com/sharepoint-connector/3.5/sharepoint-connector-reference#fileAdd

src/modules/deal-folder/dto/upload-file-in-deal-folder-request.dto.ts
```diff
@@ -0,0 +1,42 @@
...
+import { ValidatedStringApiProperty } from '@ukef/decorators/validated-string-api-property.decorator';
+
+export type UploadFileInDealFolderRequestDto = UploadFileInDealFolderRequestItem[];
+
+export class UploadFileInDealFolderRequestItem {
+  @ValidatedStringApiProperty({
+    description: 'The name of the buyer used in the deal.',
+    minLength: 1,
+    maxLength: 250,
+    pattern: /^[\w\-.()\s]+/,
```
oscar-richardson-softwire: @avaitonis Sharepoint automatically strips folder and files names of leading or trailing spaces and trailing full stops `.`, causing the querying the list for the newly created file step to fail, as the actual name of the created file path doesn't match what's expected. I see 2 possible solutions:

1. Disallow these character combinations in the request
2. Strip leading or trailing spaces and trailing full stops from what we expect the file path to be before querying the list

Which do you think is best @avaitonis?

src/modules/graph/graph.service.ts
avaitonis: Agree that in Mulesoft it is not overriding file, I don't remember this being an issue.

I could vote for renaming file with N+1th. Fail is also fine as it should highlight issues.

@abhi-markan what do you think should happen if file exists in Sharepoint?

src/modules/deal-folder/dto/upload-file-in-deal-folder-request.dto.ts
avaitonis: Both options work for me. @abhi-markan  what is your preference? maybe Portal has some filename validation?

src/modules/deal-folder/deal-folder.service.ts
```diff
@@ -0,0 +1,191 @@
...
+    const fileSizeInBytes = await this.checkFileIsNotTooLargeAndReturnSizeInBytes(fileName, fileLocationPath);
+
+    const file = await this.downloadFileFromDtfs(fileName, fileLocationPath);
+
+    await this.uploadFileToSharepoint(file, fileSizeInBytes, fileName, dealId, buyerName, ukefSiteId);
+
+    const encodedFullFilePath = await this.updateFileInformationInSharepoint(fileName, dealId, buyerName, ukefSiteId, documentType);
+
+    return {
+      fileUpload: encodedFullFilePath,
```
oscar-richardson-softwire: @avaitonis The Mulesoft code returns the `FileRef` property on the list item with the information about the newly uploaded file. However, because Mulesoft uses the Sharepoint REST API (via the Sharepoint Connector), whereas the Node version of eStore uses the Microsoft Graph API, not all the properties are the same, and indeed the `FileRef` property doesn't appear to exist on Microsoft Graph's list items.

I'm currently returning the encoded (e.g., with spaces replaced by %20) form of the file path, e.g., `Buyer%20Name/D%200030000322/fileName.fileExtension`. I just wanted to check that this is the same thing as what Mulesoft returns?

lmarrai-sw: @oscar-richardson-softwire, @avaitonis is going to check what the response looks like on Mulesoft

We had a look at DTFS and noticed it doesn't use the response (other than to log it) so if we're having trouble returning the same thing that Mulesoft returns then we could suggest returning something else and check with @abhi-markan if that'd still be good for DTFS

src/modules/deal-folder/dto/upload-file-in-deal-folder-request.dto.ts
lmarrai-sw: (1) sounds good to me, but would be good to know what @abhi-markan thinks too

src/modules/graph/graph.service.ts
IgnatG: We shouldn't override the file if it already exists. this is because Underwriters could make changes to a file with a similar name in SP and we could override their work. 

Let's use the 400 error - this is because I wouldn't want to have a lot of files with similar names to exist

oscar-richardson-softwire: Thanks @IgnatG, I've made this change

src/modules/deal-folder/deal-folder.service.ts
avaitonis: From Mulesoft logs I can see that old/current response is single object
{
  "fileUpload": "/sites/00703107/CaseLibrary/Exporter_name_example/D 0030123456/file_1.docx"
}

src/modules/deal-folder/dto/upload-file-in-deal-folder-request.dto.ts
abhi-markan: We don't have any file name validation (as far I am aware) in DTFS, perhaps best we go with option 1 and as part of the DTFS migration to eStore we can add these restriction at the UI level.

IgnatG: we don't have a file validation, but we do strip (as much as possible) special characters from a file. This is what DTFS uses to format the file names for SP 
```js
const formatFilenameForSharepoint = (filenameWithExtension) => {
  const [extension] = filenameWithExtension.match(/\.[^/.]+/g).reverse();
  const filenameWithoutExtension = filenameWithExtension.replace(/\.[^/.]+$/, '');
  const sanitisedFilename = filenameWithoutExtension.replace(/[\W_]+/g, '_');

  return `${sanitisedFilename}${extension}`;
};
```

oscar-richardson-softwire: I've added regex to implement 1) just to be safe

src/modules/deal-folder/deal-folder.service.ts
oscar-richardson-softwire: I've changed the format of the response to match this

lmarrai-sw (approved):
LGMT thanks @oscar-richardson-softwire

PR #93, opened 2023-07-05
feat(APIM-139): migrate post create facility folder endpoint

AlexBramhill:
# Introduction
This ticket migrates the POST `/site/{sitename}/deal/{dealid}/facility endpoint.

The body for this request is:
```
{
  exporterName
  buyerName
  facilityIdentifier
  destinationMarket
  riskMarket
}
```

# Resolution
This ticket makes two calls via graph api - One to gather parent folder id, and the other to gather facilityGUID. Both of these are handled as `400` inside estore, with a message explaining the issue.
It then makes a custodian folder creation request -- any error here is handled as a `500` in estore.
On success, it returns a `201` with the folder name

# Misc
I've taken some active steps to refactor some code here, for instance creation of a `tfisSharepointUrl`. The goal here is to make code more readable when making graph queries.

Some outstanding todos:
- [x] There are new env vars to add (including one that is renamed)
- [ ] facilityListId appears to be pointing at the deal folder, not facility folder.
- [x] Remove destinationMarket and riskMarket from the request -- these are not used
- [x] Add query test helpers
- [x] Unit tests have been discussed with Luke, and will be taken from his ticket

It's worth noting that the `siteId` validation parameters are more restrictive than the ticket, and are in line with our usual `siteId` validations -- I'd appreciate confirmation that this is okay.

PR #99, opened 2023-07-06
feat(APIM-136): add endpoint to create deal folder

lmarrai-sw:
## Introduction
The `POST /sites/{siteId}/deals` endpoint allows the user to create a deal folder in Sharepoint.

The request body is an array of one object with the following fields:
- exporterName
- buyerName
- dealIdentifier
- destinationMarket
- riskMarket

(see the ticket for the validation rules).

The response body is an object with the following fields: 
- folderName

## Resolution

I've added a `SharepointService` to commonise list item queries and a `DealFolderService` for creating the deal folder.

## Misc

I still need to integrate this with #93 so there a good number of TODOs to discuss with @AlexBramhill


src/config/sharepoint.config.ts
```diff
@@ -4,20 +4,24 @@ export interface SharepointConfig {
...
+  tfisCaseSitesListId: string;
+  taxonomyHiddenListTermStoreListId: string;
 }
 
 // TODO: apim-139: raise ticket for deployment of new env vars
 // TODO: apim-139: Check facility list id is named correctly (and not deal id)
 export default registerAs('sharepoint', (): SharepointConfig => {
   return {
-    tfisSharepointUrl: `sites/${process.env.SHAREPOINT_MAIN_SITE_NAME}.sharepoint.com:/sites/${process.env.SHAREPOINT_TFIS_SITE_NAME}`,
+    tfisSharepointUrl: `sites/${process.env.SHAREPOINT_MAIN_SITE_NAME}.sharepoint.com:/sites/${process.env.SHAREPOINT_TFIS_SITE_NAME}`, // TODO APIM-136: Move : into here
```
AlexBramhill: todo reminder

src/constants/examples.constant.ts
```diff
@@ -10,4 +10,7 @@ export const EXAMPLES = {
   FILE_NAME: process.env.DTFS_STORAGE_EXAMPLE_FILE_NAME,
   FILE_LOCATION_PATH: process.env.DTFS_STORAGE_EXAMPLE_FILE_LOCATION_PATH,
   FOLDER_NAME: 'Example folder name', // TODO apim-139: Check this folder name
+  BUYER_NAME: 'Example Buyer',
+  EXPORTER_NAME: 'Example Exporter',
+  MARKET: 'Germany',
```
AlexBramhill, line 15: I don't know if you've checked that this is a valid `market` -- if you were taking from my previous ticket where I used `Germany` -- I had not checked it (but since it was removed, I didn't end up finding whether `Germany` is a valid `market`

src/modules/graph/dto/common/graph-list-item-fields.dto.ts
```diff
@@ -1,8 +1 @@
-export interface GraphListItemFields {
-  Title?: string;
-  ServerRelativeUrl?: string;
-  Code?: string;
-  id: string;
-  ParentCode?: string;
-  FacilityGUID?: string;
-}
+export type GraphListItemFields = Record<string, string>; // TODO APIM-136: Can we remove this altogether?
```
AlexBramhill: Todo reminder

We...probably can? As long as we're checking the existing of each field before calling it -- which judging by the facility folder ticket, it appears to be the case.

src/modules/sharepoint/list-item-filter/and.list-item-filter.ts
```diff
@@ -0,0 +1,9 @@
+import { ListItemFilter } from './list-item-filter.interface';
+
+export class AndListItemFilter {
+  constructor(private readonly firstFilter: ListItemFilter, private readonly secondFilter: ListItemFilter) {}
```
AlexBramhill, line 4: This is great -- I feel if this was going to be a larger code base, I'd be tempted to suggest using `.join` and taking an array of items to join, but its really not needed here

src/modules/sharepoint/list-item-filter/field-equals.list-item-filter.ts
```diff
@@ -0,0 +1,13 @@
+import { ListItemFilter } from './list-item-filter.interface';
```
AlexBramhill, line 1: Looking at the file name -- are `fields` only on `list-item` or can this folder be renamed?

src/modules/site-deal/deal-folder.service.test.ts
```diff
@@ -0,0 +1,353 @@
+import { SharepointService } from '@ukef/modules/sharepoint/sharepoint.service';
+import { CreateDealFolderGenerator } from '@ukef-test/support/generator/create-deal-folder-generator';
+import { RandomValueGenerator } from '@ukef-test/support/generator/random-value-generator';
+import { when, WhenMockWithMatchers } from 'jest-when';
+
+import { CustodianService } from '../custodian/custodian.service';
```
AlexBramhill, line 6: Odd these haven't changed to `@ukef`?

src/modules/sharepoint/list-item-filter/field-equals.list-item-filter.ts
lmarrai-sw: Yes, fields are a listItem thing rather than a general Microsoft Graph thing :)

src/constants/examples.constant.ts
lmarrai-sw: Markets are country names, so Germany is a valid example (whether or not it matches any of the markets in Sharepoint) :)

avaitonis (approved):
LGTM

PR #107, opened 2023-07-10
fix(APIM-531): fix swagger regex

AlexBramhill:
# Introduction

nestjs/swagger decorator `apiProperty` takes a string for `pattern`, and not a Regexp. As a result, we turn our existing regexp into strings to be used in swagger.

In the old implementation of this, regex expressions were getting split prematurely if there was an additional `/` in the expression. 

# Solution

I've created a helper function to avoid this occurring, as apposed to relying on `split`


src/helpers/regex.helper.test.ts
```diff
@@ -0,0 +1,33 @@
...
+      const regexString = regexToSting(regex);
+
+      expect(regexString).toBe('test');
+    });
+
+    it('escapes regexp forward slashes inside a string', () => {
+      const regex = /test\/test/;
+      const regexString = regexToSting(regex);
+
+      expect(regexString).toBe('test\\/test');
```
AlexBramhill, line 17: I'm a little tempted in these tests to convert the string back to regexp and assert that the `regex->string->regex` is equal to the initial `regex`. Happy to stick or change, depending on what reviewers think

AlexBramhill: I've updated to include both these assertions

lmarrai-sw (dismissed):
LGTM thanks @AlexBramhill

lmarrai-sw (approved):

PR #113, opened 2023-07-11
feat(APIM-143): add endpoint to create buyer folders

lmarrai-sw:
## Introduction

The `POST /sites/{siteId}/buyers` endpoint allows the user to create a buyer folder in Sharepoint.

The request body is an array of one object with the following fields:

exporterName
buyerName

The response body is an object with the following fields:

buyerName

## Resolution

@francescastocco has created a `site-buyer` module for creating the buyer folder for a site.

The Mulesoft code makes 2 queries to the `tfisCaseSitesList`: once to match `URL eq {siteId}` and once to match `Title eq {exporterName}`. Both of those queries should match the same list item, so @francescastocco has simplified that into 1 query.

## Misc
The `SiteURL` field on the `tfisCaseSitesList` has a object value (rather than a string value, like all the fields we've used before have). @francescastocco has made the `ListItem` type more flexible so the type system can handle fields with _any_ type rather than only strings.


src/modules/site-buyer/site-buyer.service.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  private async getBuyerTermFromList(siteId: string, exporterName: string) {
+    const searchResults = await this.sharepointService.findListItems<{ TermGuid: string; Title: string; URL: string; SiteURL: { Url: string } }>({
+      siteUrl: this.sharepointConfig.tfisSharepointUrl,
+      listId: this.sharepointConfig.tfisCaseSitesListId,
+      fieldsToReturn: ['TermGuid', 'Title', 'URL', 'SiteURL'],
+      filter: new FieldEqualsListItemFilter({ fieldName: 'Title', targetValue: exporterName }),
+    });
+
+    if (searchResults.length === 0) {
```
IgnatG: can we use `if (!searchResults.length) {...}`

.github/workflows/deployment.yml
```diff
@@ -133,11 +133,12 @@ jobs:
             "SHAREPOINT_MAIN_SITE_NAME=${{ secrets.SHAREPOINT_MAIN_SITE_NAME }}" \
             "SHAREPOINT_TFIS_SITE_NAME=${{ secrets.SHAREPOINT_TFIS_SITE_NAME }}" \
             "SHAREPOINT_SC_SITE_NAME=${{ secrets.SHAREPOINT_SC_SITE_NAME }}" \
-            "SHAREPOINT_TFIS_FACILITY_LIST_ID=${{ secrets.SHAREPOINT_TFIS_FACILITY_LIST_ID }}" \
-            "SHAREPOINT_TFIS_FACILITY_HIDDEN_LIST_TERM_STORE_ID=${{ secrets.SHAREPOINT_TFIS_FACILITY_HIDDEN_LIST_TERM_STORE_ID }}" \
-            "SHAREPOINT_TFIS_DEAL_LIST_ID=${{ secrets.SHAREPOINT_TFIS_DEAL_LIST_ID }}" \
-            "SHAREPOINT_TFIS_CASE_SITES_LIST_ID=${{ secrets.SHAREPOINT_TFIS_CASE_SITES_LIST_ID }}" \
-            "SHAREPOINT_TAXONOMY_HIDDEN_LIST_TERM_STORE_LIST_ID=${{ secrets.SHAREPOINT_TAXONOMY_HIDDEN_LIST_TERM_STORE_LIST_ID }}" \
+            "SHAREPOINT_SC_CASE_SITES_LIST_ID=${{ secrets.SHAREPOINT_SC_CASE_SITES_LIST_ID }}" \
+            "SHAREPOINT_TFIS_FACILITY_LIST_ID: ${{ secrets.SHAREPOINT_TFIS_FACILITY_LIST_ID }}" \
```
IgnatG: should this be `=` rather than `:`?
i.e.
```yml
"SHAREPOINT_TFIS_FACILITY_LIST_ID=${{ secrets.SHAREPOINT_TFIS_FACILITY_LIST_ID }}" \
```

IgnatG: same for the other items below

src/modules/site-buyer/site-buyer.service.ts
```diff
@@ -0,0 +1,130 @@
...
+      id: string;
+      CustodianSiteURL: string;
+    }>({
+      siteUrl: this.sharepointConfig.scSharepointUrl,
+      listId: this.sharepointConfig.scCaseSitesListId,
+      fieldsToReturn: ['id', 'CustodianSiteURL'],
+      filter: new FieldEqualsListItemFilter({ fieldName: 'CustodianSiteURL', targetValue: siteId }),
+    });
+
+    if (searchResults.length === 0) {
```
IgnatG: same as the other comment, we can simplify this into `if (!searchResults.length) { ... }`

IgnatG (changes requested):
Good work! some small changes are required, particularly in the deployment.yml file

src/modules/site-buyer/dto/create-buyer-folder-params.dto.ts
```diff
@@ -0,0 +1,8 @@
+import { ValidatedSiteIdentifierApiProperty } from '@ukef/decorators/validated-site-identifier-api-property';
+import { UkefSiteId } from '@ukef/helpers';
+
+export class CreateBuyerFolderParamsDto {
+  // TODO apim-139: This is more restrictive than the ticket.
+  @ValidatedSiteIdentifierApiProperty({ description: 'The site name/identifier returned when the site was created by POST /sites.' })
+  siteId: UkefSiteId;
+}
```
lmarrai-sw: To match #115, we should:
- update the ticket with the new siteId validation and remove the TODO
- update the description of the field to match #115
- remove all references to `UkefSiteId` in the code

src/modules/site-buyer/exception/site-exporter-invalid.exception.test.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  it('exposes the name of the exception', () => {
+    const exception = new SiteExporterInvalidException(message);
+
+    expect(exception.name).toBe('SiteExporterInvalidException');
+  });
+
+  it('is instance of Error', () => {
+    expect(new SiteExporterInvalidException(message)).toBeInstanceOf(Error);
+  });
```
lmarrai-sw: Should this actually be an instance of `SiteBuyerException`?

lmarrai-sw (commented):
I've done a sweep of the code myself and found the following to change:

src/modules/site-buyer/exception/site-exporter-invalid.exception.test.ts
lmarrai-sw: I've made this change

IgnatG (approved):
LGTM

PR #114, opened 2023-07-11
fix(APIM-136): correct ne operator in list item filter

lmarrai-sw:
## Introduction

#99 refactored #93 and accidentally changed the `ne null` filter to `neq null`. This should have been caught by our API tests but the assertions don't check the exact value of the filter we use - I'll raise a ticket to improve that.

## Resolution

I've changed `neq` back to `ne` and checked this works.


IgnatG (approved):
LGTM

avaitonis (approved):
LGTM

PR #115, opened 2023-07-11
fix(APIM-533): support custom site ids

lmarrai-sw:
## Introduction

Previously our validation has assumed that `siteId` is always a number generated by the number generator. However, sometimes the site ID will be a manually created string that does not follow the same pattern. We therefore need to relax our `siteId` validation.

## Resolution

The old `siteId` validation was:
- string
- required
- length: 8
- pattern: `^0070\d{4}$`

and the new validation is 
- string
- required
- minLength: 1

I changed the code to implement this new validation, then made sure the API tests were testing this properly (it's the first time we've had a field that validates `minLength` but not `maxLength`, so I had to make some changes to support this).

I changed the documentation for `siteId` to make it more consistent with `dealId`, added a comment about the existence of custom `siteId`s that aren't from the number generator, and removed the `UkefSiteId` type as it was no longer useful.

## Misc

I removed the `temporary-file-check` module because it is no longer required (we only added it as a step towards creating `POST /sites/{siteId}/deals/{dealId}/documents`)


src/modules/site/site.service.ts
```diff
@@ -69,7 +68,7 @@ export class SiteService {
 
     const status = convertToEnum<typeof SiteStatusEnum>(siteStatus, SiteStatusEnum);
 
-    return { siteId: siteId as UkefSiteId, status };
+    return { siteId: siteId, status };
```
IgnatG: this can be simplified into
```js
return { siteId, status };
```

src/modules/site/site.service.ts
```diff
@@ -87,7 +86,7 @@ export class SiteService {
...
```
IgnatG: same as above

IgnatG (approved):
LGTM

PR #117, opened 2023-07-12
feat(apim-483): add validation for the file type in the document upload endpoint

francescastocco:
## Introduction
ESTORE should validate the file extension of files being uploaded to Sharepoint to prevent executables being uploaded, DTFS only allows the following file types BMP, DOC, DOCX, GIF, JPEG, JPG, MSG, PDF, PNG, PPT, PPTX, TIF, TXT, XLS, XLSX or ZIP

## Resolution
- Create a constant of the file types (same as the DTFS list) so it's simple to see which file types are allowed
- Used the constant in the regex for validating `fileName` on the POST /sites/{siteId}/deals/{dealId}/documents endpoint
- Made the regex case insensitive to allow for eg. '.doc', '.Doc', '.DOC' etc.

## To Do
- Add tests for this validation

## Misc
- Added some environment variables to deployment.yml, test.yml and docker-compose.yml


src/constants/document-file-type.constant.ts
```diff
@@ -0,0 +1,18 @@
+export const DOCUMENT_FILE_TYPE = {
```
lmarrai-sw: `ALLOWED_DOCUMENT_FILE_TYPE` might be a bit better to emphasise that changing this object changes the document types that the client is allowed to use (rather than just being for any file type you want to include in the code) - what do you think?

src/constants/sharepoint.constant.ts
```diff
@@ -1,5 +1,10 @@
+import { DOCUMENT_FILE_TYPE } from './document-file-type.constant';
+
 export const SHAREPOINT = {
   RESOURCE_NAME: {
     REGEX: /^[\w\-.()\s]+$/,
   },
+  DOCUMENT_FILE_NAME: {
+    REGEX: new RegExp(`^(?!\\s)[\\w\\-.()\\s]+\.(${Object.values(DOCUMENT_FILE_TYPE).join('|')})(?<![\\s.])$`, 'i'),
```
lmarrai-sw: GitHub is commenting `Found non-literal argument to RegExp Constructor` on this line, but we should be fine here because there is no user input used in the RegExp

lmarrai-sw (commented):
Thanks for this @francescastocco

src/constants/document-file-type.constant.ts
francescastocco: Yes, that sounds like a good idea

src/constants/sharepoint.constant.ts
```diff
@@ -1,5 +1,10 @@
+import { ALLOWED_DOCUMENT_FILE_TYPE } from './allowed-document-file-type.constant';
+
 export const SHAREPOINT = {
   RESOURCE_NAME: {
     REGEX: /^[\w\-.()\s]+$/,
   },
+  DOCUMENT_FILE_NAME: {
+    REGEX: new RegExp(`^(?!\\s)[\\w\\-.()\\s]+\.(${Object.values(ALLOWED_DOCUMENT_FILE_TYPE).join('|')})(?<![\\s.])$`, 'i'),
```
francescastocco: [OpenAPI doesn't appear to support case-insensitive patterns](https://stackoverflow.com/a/60773255) so we will remove this from the regex. The validation will then match DTFS and the OpenAPI documentation will be accurate

lmarrai-sw (approved):
LGTM, thanks @francescastocco!

PR #124, opened 2023-07-14
feat(APIM-474): update exporterName handling

AlexBramhill:
# Introduction

We currently have strict exporter name validation in eStore. This tickets relax this to the standards here: https://support.microsoft.com/en-gb/office/restrictions-and-limitations-in-onedrive-and-sharepoint-64883a5d-228e-48f5-b3d2-eb39e07630fa

# Resolution
- Resolution achieved using regex.
- I've also updated several tests to use new test helpers.

# Misc
- Some exporterNames are now more restrictive than previously. These are marked with TODOs
- The generator has a very rare chance to be flakey (marked with a todo).
- The regex now passes redos checkers


src/constants/exporter-name.constant.ts
```diff
@@ -0,0 +1,38 @@
...
+
+// TODO apim-474 resolve Found non-literal argument to RegExp constructor
+const disallowedPrefixes = ['\\s', '~\\$'];
+const disallowedSuffixes = ['\\s'];
+const disallowedCharacters = ['"', '*', '<', '>', '?', '/', '\\\\', '\\|'];
+const disallowedSubstrings = ['_vti_'];
+const disallowedStrings = ['.lock', 'CON', 'PRN', 'AUX', 'NUL', 'COM[0-9]', 'LPT[0-9]'];
+const disallowedStringsOnRoot = ['forms'];
+
+const disallowedPrefixesRegex = new RegExp(`(?!^(?=${disallowedPrefixes.join('|')}).*$)`);
```
AlexBramhill: This is a potential redos risk and needs to be investigated

AlexBramhill: updated

AlexBramhill:
Final regex is
`/^(?!(?=(\s)|(~\$).*$))(?!.*(\s)$)(?!.*["*<>?/\\\|])(?!.*(_vti_))(?!^(.lock|CON|PRN|AUX|NUL|COM\d|LPT\d)$)(?!(forms)$).*$/`

test/support/generator/random-value-generator.ts
```diff
@@ -29,10 +29,15 @@ export class RandomValueGenerator {
     return this.chance.string({ length, pool: '0123456789' });
   }
 
+  // TODO apim-474 This has the potential to be flakey. Realistically - We should never get so unlucky as to trigger the loop
```
AlexBramhill: I could avoid this by reducing the letter pool down a touch, as we already are testing this regex elsewhere. Will do that shortly

src/modules/site-buyer/dto/create-buyer-folder-request.dto.ts
```diff
@@ -1,10 +1,12 @@
 import { EXAMPLES } from '@ukef/constants';
+import { ValidatedExporterNameApiProperty } from '@ukef/decorators/validated-exporter-name-api-property';
 import { ValidatedSharepointResourceNameApiProperty } from '@ukef/decorators/validated-sharepoint-resource-name-api-property';
 
 export type CreateBuyerFolderRequestDto = CreateBuyerFolderRequestItem[];
 
 export class CreateBuyerFolderRequestItem {
-  @ValidatedSharepointResourceNameApiProperty({ description: 'The name of the exporter used in the deal.', example: EXAMPLES.EXPORTER_NAME })
+  // TODO apim-474 this property is now more restrictive with length
```
lmarrai-sw: What's the plan with these TODOs? Is it to record the affected endpoints on the ticket so we can have them re-tested?

lmarrai-sw: For what it's worth, I don't think the _length_ validation has changed on any of the `exporterName` fields, only the regex

test/site/post-site.api-test.ts
```diff
@@ -166,87 +166,36 @@ describe('createSite', () => {
...
-        .mockSuccessfulGraphGetCall({ value: [] });
+      mockGraphClientService.mockSuccessfulGraphApiCall().mockSuccessfulExpandCall().mockSuccessfulFilterCall().mockSuccessfulGraphGetCall({ value: [] });
 
       mockMdmService.mockSuccessfulReturnValue(siteId);
 
-      mockGraphClientService
-        .mockSuccessfulGraphApiCallWithPath(graphServicePostParams[0].path)
-        .mockSuccessfulGraphPostCallWithRequestBody(expect.anything(), graphCreateSiteResponseDto[0]);
+      mockGraphClientService.mockSuccessfulGraphApiCall().mockSuccessfulGraphPostCall(graphCreateSiteResponseDto[0]);
     };
```
lmarrai-sw: Why did we stop specifying the path, expand string, and filter string here?

src/constants/exporter-name.constant.test.ts
```diff
@@ -0,0 +1,56 @@
...
+  disallowedSubstringTestCases,
+  disallowedSuffixTestCases,
+} from '@ukef-test/common-test-cases/exporter-name-test-cases';
+
+import { EXPORTER_NAME } from './exporter-name.constant';
+
+describe('ExporterName', () => {
+  describe('Regex', () => {
+    describe('Strings', () => {
+      it.each(allowedStringTestCases)('$testTitle', ({ value }) => {
```
lmarrai-sw: These test names don't look like they'll end up being very descriptive, e.g.

```
FAIL ExporterName > Regex > Strings > Disallowed as a string: .lock

in src/constants/exporter-name.constant.test.ts
```

Could we change the `describe` titles and/or add something before `$testTitle` to make it more obvious what we're testing and what's gone wrong if it has failed, please? :)

lmarrai-sw: Also, just a small thing: we usually start our `describe`/`it` titles with lower case unless it's the name of a class / function / constant (in which case we match the capitalisation of that _thing_) - could we keep to that here as well please?

src/constants/exporter-name.constant.ts
```diff
@@ -0,0 +1,37 @@
...
+    [
+      regexToString(disallowedPrefixesRegex),
+      regexToString(disallowedSuffixesRegex),
+      regexToString(disallowedCharactersRegex),
+      regexToString(disallowedSubstringsRegex),
+      regexToString(disallowedStringsRegex),
+      regexToString(disallowedStringsOnRootRegex),
+    ].join('') +
+    '.*$',
+);
```
lmarrai-sw, line 33: Why do we:
- define each one as a regex
- turn each of the regexes into a string
- turn the combined string into a regex

instead of defining each of them as strings to begin with?

src/constants/exporter-name.constant.ts
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+const disallowedPrefixes = ['\\s', '~\\$'];
+const disallowedSuffixes = ['\\s'];
+const disallowedCharacters = ['"', '*', '<', '>', '?', '/', '\\\\', '\\|'];
+const disallowedSubstrings = ['_vti_'];
+const disallowedStrings = ['.lock', 'CON', 'PRN', 'AUX', 'NUL', 'COM\\d', 'LPT\\d'];
+const disallowedStringsOnRoot = ['forms'];
+
+const disallowedPrefixesRegex = new RegExp(`(?!(?=(${disallowedPrefixes.join(`)|(`)}).*$))`);
```
lmarrai-sw: Do we need:
- the positive lookahead wrapper `(?= ... )`
   - the reason I ask is that the negative lookahead should already be making sure we don't "consume" characters, do we need the extra level of lookahead for some reason?
- the `.*$`
   - the reason I ask is that if we're only checking the prefix, do we need to know that the string satisfies `.*` to the end?

src/constants/exporter-name.constant.ts
```diff
@@ -0,0 +1,37 @@
...
+const disallowedSuffixesRegex = new RegExp(`(?!.*(${disallowedSuffixes.join(')|(')})$)`);
+const disallowedCharactersRegex = new RegExp(`(?!.*[${disallowedCharacters.join('')}])`);
+const disallowedSubstringsRegex = new RegExp(`(?!.*(${disallowedSubstrings.join(')|(')}))`);
+const disallowedStringsRegex = new RegExp(`(?!^(${disallowedStrings.join('|')})$)`);
```
lmarrai-sw: Should the join here be `)|(` instead?

src/constants/exporter-name.constant.ts
```diff
@@ -0,0 +1,37 @@
...
+const disallowedStringsOnRootRegex = new RegExp(`(?!(${disallowedStringsOnRoot.join(')|(')})$)`);
```
lmarrai-sw: Is there a reason that `disallowedStrings` has `^` and `disallowedStringsOnRoot` does not?

test/common-test-cases/exporter-name-test-cases.ts
```diff
@@ -0,0 +1,72 @@
+interface exporterNameRegexTestCase {
```
lmarrai-sw: Can we use UpperCamelCase for interfaces please?

test/common-tests/request-field-validation-api-tests/exporter-name-field-validation-api-tests.ts
```diff
@@ -0,0 +1,84 @@
...
+    beforeEach(() => {
+      givenAnyRequestBodyWouldSucceed();
+    });
+
+    const requestIsAnArray = Array.isArray(validRequestBody);
+    const requestBodyItem = requestIsAnArray ? validRequestBody[0] : validRequestBody;
+    const fieldNameToUpdate = fieldName.toString();
+
+    it.each([...allowedStringTestCases, ...allowedSubstringTestCases, ...allowedPrefixTestCases, ...allowedSuffixTestCases])(
+      `returns a ${successStatusCode} response if exporterName matchs the regular expression ${EXPORTER_NAME.REGEX} ($testTitle)`,
```
lmarrai-sw: small typo here: `matches` instead of `matchs`

src/constants/exporter-name.constant.ts
AlexBramhill: I did this to make it easier to maintain -- I felt it's easier reading the regex as regex, without all the escape characters...Whats your opinion on this?

src/modules/site-buyer/dto/create-buyer-folder-request.dto.ts
AlexBramhill: removed these now

test/site/post-site.api-test.ts
AlexBramhill: As this is part of `givenAnyRequestBodyWouldSucceed`, we can not be sure what it is hitting `graphClientService` with -- The path, filters, or expansion could be different given a different initial request.

src/constants/exporter-name.constant.ts
AlexBramhill: Having this as suggested fails the tests -- it seems to be detecting substring if changed as above (ive updated the rootRegex to just be `|`)

src/constants/exporter-name.constant.ts
AlexBramhill: This is pretty odd -- Its not failing without the `^`, but will fail without the `$`. I'm temped to keep both in for clarity?

AlexBramhill: removed redundant one

francescastocco:
> Final regex is `/^(?!(?=(\s)|(~\$).*$))(?!.*(\s)$)(?!.*["*<>?/\\\|])(?!.*(_vti_))(?!^(.lock|CON|PRN|AUX|NUL|COM\d|LPT\d)$)(?!(forms)$).*$/`

I put this into https://regex101.com/ and empty string matched, I assume we don't want that for exporter names

src/constants/exporter-name.constant.ts
lmarrai-sw: That makes sense, happy to leave as-is

lmarrai-sw (approved):
LGTM, thanks @AlexBramhill

PR #138, opened 2023-07-19
feat(APIM-528): add API tests for upload file in deal folder endpoint

oscar-richardson-softwire:
## Introduction

When we completed [APIM-138](https://ukef-dtfs.atlassian.net/browse/APIM-138) and added the endpoint to upload files to SharePoint, we didn't include API or unit tests for this functionality.

## Resolution

This PR adds API (but not unit) tests for this functionality.


test/deal-folder/post-document-in-deal-folder.api-test.ts
```diff
@@ -56,22 +67,327 @@ describe('postDocumentInDealFolder', () => {
...
+      minLength: 1,
+      maxLength: 250,
+      generateFieldValueOfLength: (length: number) => valueGenerator.buyerName({ length }),
+      pattern: BUYER_NAME.REGEX,
+      generateFieldValueThatDoesNotMatchRegex: () => ' Example Buyer',
+      validRequestBody: uploadFileInDealFolderRequest,
+      successStatusCode,
+      makeRequest: (requestBody: unknown[]) => makeRequestWithBodyAndMockReturningItemId(requestBody as UploadFileInDealFolderRequestDto),
+      givenAnyRequestBodyWouldSucceed: () => givenAnyRequestWouldSucceedUpToReturningItem(),
+    });
```
lmarrai-sw: We test `buyerName` twice :)

lmarrai-sw (changes requested):
Thanks @oscar-richardson-softwire

test/deal-folder/post-document-in-deal-folder.api-test.ts
oscar-richardson-softwire: Good spot, removed :)

src/constants/file-location-path.constant.ts
```diff
@@ -1,3 +1,4 @@
 export const FILE_LOCATION_PATH = {
-  REGEX: /^[\w\-:/\\()\s]+$/,
+  // [a-fA-F\d]{24} matches a MongoDB hex, which will be the name of the first folder in the path.
+  REGEX: /^[a-fA-F\d]{24}(\/[\w\-:\\()\s]+)*$/,
```
lmarrai-sw, line 3: It looks like we've accidentally changed the `REGEX` here 

EDIT: I was wrong, we didn't change the regex, but the linter is marking it as unsafe

lmarrai-sw (changes requested):

src/constants/file-location-path.constant.ts
lmarrai-sw: The linter uses the [safe-regex](https://www.npmjs.com/package/safe-regex) package which can give false positives as it's a simple (but quick) check. I've put the regex into another checker (https://devina.io/redos-checker, which uses [recheck](https://github.com/makenowjust-labs/recheck)) which says that it's safe. I'll bring this up with the team to see if we want to move to a better but slower regex checker

lmarrai-sw (approved):

PR #142, opened 2023-07-20
feat(APIM-561): restrict upload file in deal folder endpoint to only use files from dtfs deal folder

oscar-richardson-softwire:
## Introduction

Currently the POST /sites/{siteId}/deals/{dealId}/documents endpoint can upload any file from DTFS’s Azure File Storage account to Sharepoint because the user has full control over the path. This is a concern because if a malicious party has access to both:

- an API key to use the estore API, and

- the estore Sharepoint site for an exporter

then they can upload any file (that they can guess the path for) from DTFS’s storage to Sharepoint and then access the file.

## Resolution

- Added an env variable for the path to the folder in DTFS storage for deal documents that should be uploaded to eStore
- Prefixed the `fileLocation` provided in the request with this value in the `File Share Client`
- Added a regex to ensure the `fileLocation` starts with a 24-character MongoDB hex and does not contain any period characters (`.`), as this could allow access to the parent directory via `..`. Similarly `~` is disallowed as it could allow access to the root directory
- Updated the API differences document to record this change
- Raised a [ticket](https://ukef-dtfs.atlassian.net/browse/APIM-567) for @abhi-markan to complete the necessary actions associated with the change to the `.env`
- Updated `docker-compose.yml`, `test.yml` and `deployment.yml` files.


src/modules/deal-folder/dto/upload-file-in-deal-folder-request.dto.ts
```diff
@@ -33,9 +33,9 @@ export class UploadFileInDealFolderRequestItem {
 
   @ValidatedStringApiProperty({
     description: 'The path of the location/folder in Azure storage from which the file needs to be loaded.',
-    minLength: 1,
+    minLength: 24,
     maxLength: 250,
-    pattern: /^[\w\-:/\\()\s]+$/,
+    pattern: /^[a-fA-F\d]{24}[\w\-:/\\()\s]*$/,
```
lmarrai-sw: I think we should write a comment to explain why `[a-fA-F\d]{24}` is in the pattern here, it'd be quite surprising to a reader otherwise

Also does it matter that we don't force a `/` after the first 24 characters if the string has length more than 24?

test/support/generator/random-value-generator.ts
```diff
@@ -118,8 +118,14 @@ export class RandomValueGenerator {
...
-    const pool = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-_():/\\ ';
-    return this.chance.string({ length, pool });
+    const length = options && (options.length || options.length === 0) ? options.length : this.chance.integer({ min: 24, max: 250 });
+    const poolForHex = `abcdefABCDEF0123456789`;
+    const poolForRestOfPath = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-_():/\\ ';
+    if (length < 25) {
+      return this.chance.string({ length, pool: poolForHex });
+    }
+    const hex = this.chance.string({ length: 24, pool: poolForHex });
+    const restOfPath = this.chance.string({ length: length - 24, pool: poolForRestOfPath });
```
lmarrai-sw: What happens when `length` is negative? Does it treat the length as 0?

test/support/mocks/dtfs-storage-client.service.mock.ts
```diff
@@ -12,12 +12,16 @@ class MockFileShareClient {
...
   readonly storageSharedKeyCredential: StorageSharedKeyCredential;
   fileShareClient: ShareFileClient;
   getShareFileClient: jest.Mock;
 
   constructor() {
-    (this.baseUrl = null), (this.storageSharedKeyCredential = null), (this.fileShareClient = new MockFileShareClient() as unknown as ShareFileClient);
+    (this.baseUrl = null),
+      (this.dealDocumentsFolderPath = null),
+      (this.storageSharedKeyCredential = null),
+      (this.fileShareClient = new MockFileShareClient() as unknown as ShareFileClient);
```
lmarrai-sw: I'm surprised by the brackets and commas here, should this not be
```
this.baseUrl = null;
this.dealDocumentsFolderPath = null;
// etc.
```
?

lmarrai-sw (changes requested):

test/support/generator/random-value-generator.ts
oscar-richardson-softwire: This will never happen because we return early if length is less than 25.

test/support/mocks/dtfs-storage-client.service.mock.ts
oscar-richardson-softwire: Agreed, I was copying the pattern that was already there (which I don't remember doing myself, so assumed it either came from linting or someone else's refactor which I got when I branched off of main) - have made this change

src/modules/deal-folder/dto/upload-file-in-deal-folder-request.dto.ts
oscar-richardson-softwire: Added comment explaining `[a-fA-F\d]{24}` part. Good point about starting with a `/` after the hex, I've adjusted the regex (and the generator function) to ensure that

src/modules/deal-folder/dto/upload-file-in-deal-folder-request.dto.ts
```diff
@@ -33,9 +33,10 @@ export class UploadFileInDealFolderRequestItem {
 
   @ValidatedStringApiProperty({
     description: 'The path of the location/folder in Azure storage from which the file needs to be loaded.',
-    minLength: 1,
+    minLength: 24,
     maxLength: 250,
-    pattern: /^[\w\-:/\\()\s]+$/,
+    pattern: /^[a-fA-F\d]{24}(\/[\w\-:/\\()\s]*)*$/,
+    // [a-fA-F\d]{24} matches a MongoDB hex, which will be the name of the first folder in the path.
```
lmarrai-sw: Sorry to be a pain but can we put the comment above the line it's referring to instead please? That's more typical than underneath :)

src/modules/deal-folder/dto/upload-file-in-deal-folder-request.dto.ts
```diff
@@ -33,9 +33,10 @@ export class UploadFileInDealFolderRequestItem {
 
   @ValidatedStringApiProperty({
     description: 'The path of the location/folder in Azure storage from which the file needs to be loaded.',
-    minLength: 1,
+    minLength: 24,
     maxLength: 250,
-    pattern: /^[\w\-:/\\()\s]+$/,
+    pattern: /^[a-fA-F\d]{24}(\/[\w\-:/\\()\s]*)*$/,
```
lmarrai-sw: Unfortunately the regex string `/^[a-fA-F\d]{24}(\/[\w\-:/\\()\s]*)*$/` is vulnerable to ReDOS attack according to https://devina.io/redos-checker

lmarrai-sw (changes requested):

src/modules/deal-folder/dto/upload-file-in-deal-folder-request.dto.ts
oscar-richardson-softwire: Done :)

src/modules/deal-folder/dto/upload-file-in-deal-folder-request.dto.ts
oscar-richardson-softwire: Changed to `/^[a-fA-F\d]{24}(\/[\w\-:\\()\s]+)*$/`, which is not vulnerable and also ensures the string does not end with '/' (which causes a 500 error)

lmarrai-sw (dismissed):
LGTM, thanks @oscar-richardson-softwire

lmarrai-sw (approved):

PR #143, opened 2023-07-20
fix(APIM-559): validate site id does not include invalid characters

francescastocco:
## Introduction
BUG-Returning 500 with `:siteId` as the siteId for uploading file in the deal folder in E-STORE

## Resolution
- Pattern added for site ID to not allow invalid sharepoint characters


src/constants/sharepoint.constant.ts
```diff
@@ -7,4 +7,7 @@ export const SHAREPOINT = {
   DOCUMENT_FILE_NAME: {
     REGEX: new RegExp(`^(?!\\s)[\\w\\-.()\\s]+\.(${Object.values(ALLOWED_DOCUMENT_FILE_TYPE).join('|')})(?<![\\s.])$`),
   },
+  CASE_SITE_URL: {
+    REGEX: /^[^,&!@~#$%*:;()+./\\[\]\-]*$/,
```
lmarrai-sw: Minor point but we could use `+` instead of `*` since it must have at least 1 character

lmarrai-sw (dismissed):
I added a minor point, aside from that it LGTM

lmarrai-sw (approved):
LGTM, thanks @francescastocco

PR #145, opened 2023-07-21
feat(APIM-538): add environment variable for log format

francescastocco:
## Introduction
Currently, all our log messages use the single line format which makes them hard to read during local development but is good for the dev, staging, and production environments.

## Resolution 
* Added an environment variable that controls if we use single line or multi-line logging format
* Use this to set the singleLine property in our logging config
* Added environment variable to `docker-compose.yml`, `test.yml` and `deployment.yml`.
* Updated `.env.sample`.

## Miscellaneous
* Dependencies updates


src/main.module.ts
```diff
@@ -22,7 +22,7 @@ import { LoggerModule } from 'nestjs-pino';
           transport: {
             target: 'pino-pretty',
             options: {
-              singleLine: true,
+              singleLine: config.get<string>('app.singleLineLogFormat'),
```
lmarrai-sw: I think this should be `config.get<boolean>` instead

lmarrai-sw (dismissed):
thanks @francescastocco, just a small change to make

lmarrai-sw (approved):

PR #147, opened 2023-07-21
feat(APIM-472): update graph service endpoint patterns

AlexBramhill:
# Introduction
We currently have REST methods exposed on GraphService. We then look to call these REST methods on applicable services - ie `SiteService.getSiteStatusByExporterName` calls `GraphService.get(<params>)`. 

For other services however, such as MdmService, we have specific and non-generic methods listed on the service (`MdmService.createNumbers`), that call private generic methods (`MdmService.post`).

We should look to move GraphService to follow this pattern - eg making GraphService.get a private method, and creating a new method GraphService.getSiteStatusByExporterNamefor each specific sharepoint endpoint.

The will allow us to limit the injection of SharepointConfig to just the GraphService.
In other services calling 3rd party endpoints, 

# Resolution

Due to the larger amount of endpoints hitting GraphService, I have chosen to utilise the existing `sharepointService`. SharepointService contains these intermediately methods between a specific service and the graph service's REST methods.

# Misc

There are still some SharePoint config imports in the main services, however these are usually used for url construction etc, and is consistent with our implementation of custodian config.


src/modules/deal-folder/deal-folder.service.ts
```diff
@@ -6,19 +6,19 @@ import { DocumentTypeEnum } from '@ukef/constants/enums/document-type';
 import { SharepointResourceTypeEnum } from '@ukef/constants/enums/sharepoint-resource-type';
 import { DocumentTypeMapper } from '@ukef/modules/deal-folder/document-type-mapper';
 import { DtfsStorageFileService } from '@ukef/modules/dtfs-storage/dtfs-storage-file.service';
-import GraphService from '@ukef/modules/graph/graph.service';
 
+import { SharepointService } from '../sharepoint/sharepoint.service';
 import { UploadFileInDealFolderResponseDto } from './dto/upload-file-in-deal-folder-response.dto';
-type RequiredConfigKeys = 'baseUrl' | 'ukefSharepointName' | 'estoreDocumentTypeIdFieldName';
 
+type RequiredSharepointConfigKeys = 'baseUrl' | 'estoreDocumentTypeIdFieldName' | 'ukefSharepointName';
```
AlexBramhill, line 13: We currently use these to create the url to upload to -- in keeping with other modules, I'm keeping this logic in this service.

src/modules/deal-folder/deal-folder.service.ts
```diff
@@ -95,23 +87,17 @@ export class DealFolderService {
     and driveId appears to necessary because using the ukefSiteId and drive name only appears to work when accessing drives, not items *within* drives. */
     const fileDestinationPath = `${buyerName}/D ${dealId}`;
 
-    return `${this.config.baseUrl}/sites/${sharepointSiteId}/drives/${driveId}/root:/${fileDestinationPath}/${fileName}:/createUploadSession`;
+    return `${this.sharepointConfig.baseUrl}/sites/${sharepointSiteId}/drives/${driveId}/root:/${fileDestinationPath}/${fileName}:/createUploadSession`;
```
AlexBramhill, line 90: We could move this function (and others referencing upload urls) to sharepoint service -- my only issue with that is that it's breaking from the pattern,

src/modules/graph/dto/graph-get-site-status-by-exporter-name-response.dto.ts
```diff
@@ -1,21 +0,0 @@
-import { GraphContentType } from './common/graph-content-type.dto';
-import { GraphParentReference } from './common/graph-parent-reference.dto';
-import { GraphSiteFields } from './common/graph-site-fields.dto';
-import { GraphUser } from './common/graph-user.dto';
-
-export interface GraphGetSiteStatusByExporterNameResponseDto {
```
AlexBramhill, line 6: Managed to get rid of this in favour of the list item object

src/modules/sharepoint/sharepoint.service.ts
```diff
@@ -1,24 +1,213 @@
...
 export class SharepointService {
-  constructor(private readonly graphService: GraphService) {}
+  constructor(
+    private readonly graphService: GraphService,
+    @Inject(SharepointConfig.KEY)
+    private readonly sharepointConfig: ConfigType<typeof SharepointConfig>,
+  ) {}
+
+  async getSiteByUkefSiteId(ukefSiteId: string): Promise<{ id: string }> {
+    return await this.graphService.get<{ id: string }>({
```
lmarrai-sw, line 63: I think we can make all the `return await`s in the file just `return` instead (and then remove the `async` modifier as well)

src/modules/deal-folder/deal-folder.service.ts
lmarrai-sw: It's a tricky one to place, I'm happy to leave it where it is for now :)

src/modules/sharepoint/sharepoint.service.test.ts
```diff
@@ -1,76 +1,218 @@
...
   const valueGenerator = new RandomValueGenerator();
 
-  let graphServiceGet: jest.Mock;
+  const siteId = valueGenerator.ukefSiteId();
+  const buyerName = valueGenerator.buyerName();
+  const exporterName = valueGenerator.exporterName();
+  const marketName = valueGenerator.string();
+  const dealFolderName = valueGenerator.string();
+  const facilityIdentifier = valueGenerator.facilityId();
+  const sharepointResourceType = valueGenerator.enumValue(SharepointResourceTypeEnum) as SharepointResourceTypeEnum;
```
lmarrai-sw, line 22: Minor point, but the intended usage here if you need to give the typing a kick is `valueGenerator.enumValue<SharepointResourceTypeEnum>(SharepointResourceTypeEnum);` instead of `as SharepointResourceTypeEnum`

src/modules/sharepoint/sharepoint.test-parts/mock-sharepoint-config.helper.ts
```diff
@@ -0,0 +1,22 @@
+import { RandomValueGenerator } from '@ukef-test/support/generator/random-value-generator';
+
+export const getMockSharepointConfig = () => {
```
lmarrai-sw, line 3: Can we record the return type of the function please?

lmarrai-sw (commented):
I've made a few minor comments @AlexBramhill but otherwise this is great. Thank you!

src/modules/sharepoint/sharepoint.test-parts/mock-sharepoint-config.helper.ts
lmarrai-sw: This will be solved in a later PR

src/modules/sharepoint/sharepoint.service.test.ts
lmarrai-sw: This will be solved in a later PR

src/modules/sharepoint/sharepoint.service.ts
lmarrai-sw: This will be solved in a later PR

lmarrai-sw (approved):

PR #165, opened 2023-07-25
fix(APIM-574): add desktop.ini to list of exporterName restrictions

AlexBramhill:
# Introduction
`desktop.ini` is not currently in our disallowed word list for `exporterName`

# Resolution
I've added `desktop.ini` to the disallowed word list for `exporterName` and added tests

# Misc
I've also fixed an issue with `.` being recognised as a wildcard for `.lock`


avaitonis (dismissed):
LGTM, 
but API test failed, maybe it was intermittent issue.

avaitonis (approved):
LGTM

PR #167, opened 2023-07-26
feat(APIM-468): change how integer config values are parsed

avaitonis:
### Introduction
Current way to parse environment variables and set default value doesn't allow using 0. 0 can be useful for CUSTODIAN_MAX_REDIRECTS, CUSTODIAN_TIMEOUT and similar integer environment variables.

### Resolution
This is copy of similar implementation in TFS API.
Created helper getIntConfig. It parses input string and returns integer from input or optional default value.

### Helper throws exception for these edge cases:
- input is set, but not string
- input and default values are undefined
- input is not integer in string format. For example it is float or has some non digital characters
- input number is not using decimal base

### Misc
Moved boolean environment variable test to shared test file withEnvironmentVariableParsingUnitTests


IgnatG (dismissed):

src/config/app.config.test.ts
```diff
@@ -79,57 +77,24 @@ describe('appConfig', () => {
     });
   });
 
-  describe('parsing SINGLE_LINE_LOG_FORMAT', () => {
```
AlexBramhill, line 82: We probably want this back in?

src/helpers/get-int-config.ts
```diff
@@ -0,0 +1,21 @@
+import { InvalidConfigException } from '@ukef/config/invalid-config.exception';
+
+// This helper function is used to get integer from configuration.
+export const getIntConfig = (environmentVariable: string, defaultValue?: number): number | undefined => {
+  if (typeof environmentVariable === 'undefined') {
+    if (typeof defaultValue === 'undefined') {
+      throw new InvalidConfigException(`Environment variable is missing and doesn't have default value.`);
+    }
+    return defaultValue;
```
AlexBramhill, line 9: Do we need to check that `defaultValue` is an int?

src/config/app.config.test.ts
avaitonis: I moved environment variable SINGLE_LINE_LOG_FORMAT test to helper withEnvironmentVariableParsingUnitTests

src/helpers/get-int-config.ts
avaitonis: I wanted to avoid complicating this too much. Also typescript provides some validation ensuring that `defaultValue` is number.

src/helpers/get-int-config.ts
```diff
@@ -0,0 +1,21 @@
+import { InvalidConfigException } from '@ukef/config/invalid-config.exception';
+
+// This helper function is used to get integer from configuration.
+export const getIntConfig = (environmentVariable: string, defaultValue?: number): number | undefined => {
```
AlexBramhill: Does this just return number?

src/helpers/get-int-config.helper.test.ts
```diff
@@ -0,0 +1,47 @@
+import { InvalidConfigException } from '@ukef/config/invalid-config.exception';
+
+import { getIntConfig } from './get-int-config';
+
+describe('Get integer configuration value Helper', () => {
```
AlexBramhill: `GetIntConfig helper` matches our existing patterns

src/helpers/get-int-config.ts
avaitonis: yes, number, but specifically just integer.

avaitonis: We could create type for integer, but not sure if this is required.

AlexBramhill: Sorry - I mean why are we doing `number | underfined`

src/helpers/get-int-config.helper.test.ts
avaitonis: Thanks, changed in Estore and MDM .

src/helpers/get-int-config.ts
avaitonis: Good catch, it is from older version of the function. I removed `undefined` from Estore and MDM APIs.

AlexBramhill (approved):

PR #170, opened 2023-07-27
feat(APIM-538): log format environmental variable

AlexBramhill:
# Introduction
This PR adds tests around the `app.config`

# Resolution
This adds tests via our existing test helpers, as well as adding a `getBooleanConfig` function for config files

# Misc
There is no comparable helpers in `mdm` or `tfs`, as a result, there will be no PR for these repos

### DIFF
--- a/src/config/app.config.test.ts
+++ b/src/config/app.config.test.ts
@@ -111,15 +111,18 @@ describe('appConfig', () => {
         testBoolString: 'false',
         expectedBoolean: false,
       },
-    ])('is the env variable HTTP_VERSIONING_ENABLE parsed as a $expectedBoolean boolean if HTTP_VERSIONING_ENABLE is $testBoolString', ({ testBoolString, expectedBoolean }) => {
-      replaceEnvironmentVariables({
-        HTTP_VERSIONING_ENABLE: testBoolString,
-      });
+    ])(
+      'is the env variable HTTP_VERSIONING_ENABLE parsed as a $expectedBoolean boolean if HTTP_VERSIONING_ENABLE is $testBoolString',
+      ({ testBoolString, expectedBoolean }) => {
+        replaceEnvironmentVariables({
+          HTTP_VERSIONING_ENABLE: testBoolString,
+        });
 
-      const config = appConfig();
+        const config = appConfig();
 
-      expect(config.versioning.enable).toBe(expectedBoolean);
-    });
+        expect(config.versioning.enable).toBe(expectedBoolean);
+      },
+    );
   });
 
   describe('parsing HTTP_VERSION', () => {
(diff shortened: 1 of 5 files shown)


src/config/app.config.test.ts
```diff
@@ -79,124 +77,106 @@ describe('appConfig', () => {
...
-        REDACT_LOGS: 'false',
-      });
-
-      const config = appConfig();
-
-      expect(config.redactLogs).toBe(false);
-    });
-
-    it('sets redactLogs to true if REDACT_LOGS is not specified', () => {
+  describe('parsing HTTP_VERSIONING_ENABLE', () => {
```
AlexBramhill, line 80: We have this as separate tests as it is a nested object and time constraints means we are not able to rework the existing helper implementation

test/common-tests/environment-variable-parsing-unit-tests.ts
```diff
@@ -168,45 +185,13 @@ export const withEnvironmentVariableParsingUnitTests = <ConfigUnderTest>({
...
-
-        it(`is the default value if ${environmentVariableName} is empty string`, () => {
-          process.env = {
-            [environmentVariableName]: '',
-          };
-
-          const { [configPropertyName]: configPropertyValue } = getConfig();
-
-          expect(configPropertyValue).toBe(defaultConfigValue);
-        });
```
lmarrai-sw, line 209: Why did we get rid of this test?

src/config/app.config.ts
```diff
@@ -42,7 +44,7 @@ export default registerAs('app', (): Record<string, any> => {
     port: getIntConfig(process.env.HTTP_PORT, 3001),
     apiKey: process.env.API_KEY,
     logLevel: process.env.LOG_LEVEL || 'info',
-    redactLogs: process.env.REDACT_LOGS !== 'false',
-    singleLineLogFormat: process.env.SINGLE_LINE_LOG_FORMAT !== 'false',
+    redactLogs: getBooleanConfig(process.env.REDACT_LOGS, true),
+    singleLineLogFormat: getBooleanConfig(process.env.SINGLE_LINE_LOG_FORMAT, false),
```
lmarrai-sw, line 48: Why did we change the default for `singleLineLogFormat`?

lmarrai-sw (dismissed):
thanks @AlexBramhill, just a couple small things for me

src/config/app.config.ts
AlexBramhill: Ah -- Changing back

test/common-tests/environment-variable-parsing-unit-tests.ts
AlexBramhill: We're now forcing defaults -- An empty string will throw a invalid boolean error (I've added a test case for this in get-boolean-config.helper.test`

lmarrai-sw (approved):
LGTM

PR #176, opened 2023-07-28
feat(APIM-472): update graph service endpoint patterns

AlexBramhill:
# Introduction
There were three review comments needing addressing following apim-472. These are addressed in this PR

- Record `getMockSharepointConfig` type in the function declaration
- Update `sharepointResourceType` in `sharepoint.service.test.ts` to use the generic type, as apposed to casting `as SharepointResourceTypeEnum`
- Remove unnecessary awaits in sharepoint service

# Resolution
I've updated to these comments in this PR.

I have kept the awaits in any `sharepointService` method that returns `void`. I have also kept the await in the `findListItems`.


lmarrai-sw (approved):
LGTM!

PR #180, opened 2023-07-31
feat(APIM-465): remove sensitive data from logs

avaitonis:
### Introduction
We log incoming, outgoing and third party calls, including headers and sensitive data. Some of these log messages contain secret or sensitive information (e.g. the client's secret api key header). The amount of logging that might contain sensitive data depends on the log level, but even our ERROR level logs can contain this information.

### Resolution
The `pino` logger we use has [functionality for redacting keys](https://github.com/pinojs/pino/blob/master/docs/redaction.md) in our log messages.

I've been through our log messages and identified the following occasions where we log sensitive information:
  - Incoming client request
  - MS graph errors. Successful call logging is minimal
  - Custodian API calls
  - MDM API calls
  - Azure storage API calls
  - Errors

### Misc
  - Azure storage error log contains a lot of data, 11 000+ lines, ~550KB. All of this is in attribute `operationSpec`, we don't need to log Azure spec, so I added `operationSpec` to redact list.
  - Added LoggingInterceptor to make Estore API more consistent with our other APIs

### Todo
- MS Graph API call logging is minimal. It should be fine for now, but could be made more similar to other calls.
- Maybe we need to unset `operationSpec` in Azure storage errors all the time, even if redacting is disabled.


src/constants/redact-strings.constant.ts
```diff
@@ -0,0 +1,4 @@
+export const REDACT_STRINGS = [
+  { searchValue: process.env.GRAPH_AUTHENTICATION_CLIENT_ID, replaceValue: '[Redacted]' },
```
IgnatG, line 2: Does the process.env load here without dotenv?

avaitonis: dotenv load happens in configuration, before this constant is used. process.env is global variable, so single dotenv load is enough.

I could add dotenv load here, but maybe it's excessive.

src/helpers/redact-strings-in-log-args.helper.ts
```diff
@@ -0,0 +1,31 @@
...
+      if (value) {
+        const safeValue = redactString(redactStrings, value);
+        set(args[index], path, safeValue);
+      }
+    });
+  });
+  return args;
+};
+
+const redactString = (redactStrings: { searchValue: string | RegExp; replaceValue: string }[], string: string): string => {
```
abhi-markan: Can a more meaningful name please be given to argument `string`, perhaps `message`?

abhi-markan (changes requested):
Looks good, just a minor comment to address please.

src/helpers/redact-strings-in-log-args.helper.ts
avaitonis: Thanks, changed to `logMessage`

abhi-markan (approved):
LGTM 👍🏼

PR #193, opened 2023-08-07
feat(APIM-576): relax buyers name to match exporter name rules

avaitonis:
### Introduction
In the past we had strict name rules for Sharepoint folders. We successfully relaxed exporter name rules. Buyer company also can have international characters and it's better to allow them.

### Resolution
- moved exporter name regex to more generic Sharepoint constants file
- copied exporter name decorator as buyer name decorator
- updated tests


IgnatG (approved):
LGTM

PR #234, opened 2023-08-23
feat(APIM-482): dont create folder if it already exists

avaitonis:
### Introduction
Custodian will accept multiple requests to create same folder, and will create multiple records internally and in Sharepoint lists that it is not possible to complete request.
We create 3 tiers of folders:
- Exporter (Sharepoint site, not a folder)
  - Buyer folder
    - Deal folder
      - Facility folder

### Resolution
Folder creation in Sharepoint used to take 5 seconds, but at the moment it is up to 5 min.

1. Check for older existing folders - before creating new folder we need to check if this folder doesn't already exist **(implemented in this PR)**
2. Check for `work in progress` folder creation - maybe Custodian started folder creation less than 5 minutes ago and we shouldn't send same request again **(not in this PR)**. I can thing of 3/4 solutions:
    1. Sharepoint has 3 lists for Custodian jobs, these lists get [work in progress] items, but with very big delay. `this option is non viable`
    2. Get information about pending jobs from Custodian using it's API. Tried multiple endpoints, but `I didn't find required information`.
    3. Use throttling based on endpoint and POST body, maybe in API, maybe in Azure APIM.

I looked at Custodian API spec, searched for Custodian documentation, had tour of Custodian admin interface, but was not able to find a way to get information about folders being created from Custodian.

### Misc
- At the moment we are loading Sharepoint lists by id's, but loading by title is also possible and titles might be easier to manage. I used Sharepoint list title for new list that was required.
- New environment variable SHAREPOINT_TFIS_FACILITY_LIST_TITLE.
- Noticed that variable names for Sharepoint lists might not match purpose of the lists.


IgnatG (approved):
LGTM

PR #292, opened 2023-09-13
fix (APIM-482): changed status from 400 to 201 in situation where Buyer, Deal or Facility folder exists

avaitonis:
## Introduction
In user story APIM-482 we introduced `400` Folder exists responses, but client is relaying on success code in such situation.

## Resolution
We decided to return `201` or `200` to avoid changes on the client. Final status code, at least for now, is `201`, because 200 requires more changes. `201` means that folder creation is not required, and client doesn't need to know details about this.

### Nice to have
We could add extra logging for folder exists events.


abhi-markan (approved):
LGTM 👍🏼

PR #341, opened 2023-10-06
feat (APIM-586): don't create folders if it is already is processed by Custodian

avaitonis:
### Introduction
Sometimes creation of folder in Sharepoint takes long time. For example in development environment it might take 3-7 minutes. It used to take 5 seconds, and some tests are still mentioning this short time delay.
Custodian accepts multiple requests to create same folder and returns 201. To avoid duplicate call we need to check for duplicate on our end.
Sharepoint folders are created for: *Buyer, Deal and Facilities*, we have 3 endpoint for this. Process for all 3 folders is similar, but because small differences they are not merged.

### Issues
- After reading Custodian specification it was no way to find *all* pending jobs
- Custodian API provides information about pending jobs if you know some information about this job. Issues are:
  - Recent pending Custodian jobs might not be returned. This delay might be from 1 second to 20 minutes.
  - Estore API need to know Folder Creation RequestID to get information about pending Custodian job.
  - Estore API doesn't have storage.

### Resolution
Why CacheManager was chosen for storage of RequestIDs:
   - This check is not critical, therefore in memory, not permanent, storage should be adequate
   - Not many RequestIDs need to be stored. In production we expect 3-9 folder creation RequestIDs a day.
   - RequestIDs needs to be stored for a short time. Just until folder creation is completed. 1-30 minutes.
   - We use CacheManager in TFS project.
   - No extra infrastructure is required.

CacheManager disadvantages:
  - It's not permanent. 
  - Redeployments of API would clear currently saved RequestIDs.

Folder creation journey **(new parts are in bold)**:
  1.  Check that Folder doesn’t exist in Sharepoint
     a. If folder exist we return 200 to the client and end journey.
  2.  **Check if we have previous Custodian RequestGUID for this folder**
     a. **If Custodian RequestGUID exist we check call Custodian to get Provisioning job:**
       1.  **If Custodian job has value for field “Failed“. We allow to create new Custodian job**
       2.  **If Custodian job didn’t fail we return 202 – “waiting for Custodian” to the client and end journey.**
  3.  Do POST /Create/CreateAndProvision . 
  4.  **Save RequestGUID from POST /Create/CreateAndProvision result**

3 POST endpoints are changed:
  - /sites/{{estore-site-id}}/buyers
  - /sites/{{estore-site-id}}/deals
  - /sites/{{estore-site-id}}/deals/{{estore-deal-id}}/facilities

Changes:
  - `folder exists in Sharepoint check` is moved from service to controller
  - `folder exists in Custodian check` added to controller
  - some extra logic had to be moved from service to controller. 
    - generating folder name
    - getting parent folder id from Sharepoint.
  - POST buyer endpoint - response field buyerName renamed to folderName. So it matches other folder endpoints.
  - New `status` field added to responses of all 3 endpoint. It can have 8 values.
  - Folder exists in Sharepoint response status changed from from 201 to 200.
  - Folder exists in Custodian response status changed from from 201 to 202.
  - RequestGUID is stored after POST `/Create/CreateAndProvision`
  - New Custodian call added POST `/Provisioning/JobsByRequestId`
  - New environment variable `CUSTODIAN_JOB_STORE_TTL_IN_MILLISECONDS`. 30 minutes should be good value for this, 1800000 in milliseconds.


src/helpers/get-custodian-folder-creation-cache-key.helper.test.ts
```diff
@@ -0,0 +1,30 @@
...
+import { getCustodianFolderCreationCacheKey } from './get-custodian-folder-creation-cache-key.helper';
+
+describe('GetCustodianFolderCreationCacheKey helper test', () => {
+  const valueGenerator = new RandomValueGenerator();
+
+  describe('getCustodianFolderCreationCacheKey', () => {
+    const parentFolderId = valueGenerator.nonnegativeInteger();
+
+    it('test cache key generation without reimplementing full method', () => {
+      const key1 = getCustodianFolderCreationCacheKey(parentFolderId, valueGenerator.string());
```
oscar-richardson-softwire: For the instances where we have `valueGenerator.string()` inline as the second argument, can we pull these out into variables (e.g., `folderName`, `anotherFolderName`) for readability

src/helpers/get-custodian-folder-creation-cache-key.helper.test.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import { getCustodianFolderCreationCacheKey } from './get-custodian-folder-creation-cache-key.helper';
+
+describe('GetCustodianFolderCreationCacheKey helper test', () => {
+  const valueGenerator = new RandomValueGenerator();
+
+  describe('getCustodianFolderCreationCacheKey', () => {
+    const parentFolderId = valueGenerator.nonnegativeInteger();
+
+    it('test cache key generation without reimplementing full method', () => {
```
oscar-richardson-softwire: I think this `it` block should be reworded (it could be read as the `getCustodianFolderCreationCacheKey` itself tests cache key generation, which it does not). I'd suggest something like `it('generates a cache key of the correct format')`

src/helpers/get-custodian-folder-creation-cache-key.helper.ts
```diff
@@ -0,0 +1,7 @@
+import { CUSTODIAN } from '@ukef/constants';
+import * as crypto from 'crypto';
+
+export const getCustodianFolderCreationCacheKey = (parentFolderId: number, folderName: string): string => {
```
oscar-richardson-softwire, line 4: It might be clearer if this function were renamed `generateCustodianFolderCreationCacheKey` to make it clear that it is actually creating the cache key rather than just getting it via a request, for instance

src/modules/custodian/custodian.service.test.ts
```diff
@@ -82,6 +135,31 @@ describe('CustodianService', () => {
 
       expect(httpServicePost).toHaveBeenCalledTimes(1);
       expect(httpServicePost).toHaveBeenCalledWith(...expectedHttpServicePostArgs);
+
+      expect(cacheManagerSet).toHaveBeenCalledTimes(2);
```
oscar-richardson-softwire, line 153: I would argue that this `it` case perhaps tests too much and we should maybe split this off into a different test (e.g., `it calls the set method on the cache manager correctly`)

src/modules/custodian/custodian.service.test.ts
```diff
@@ -93,8 +171,160 @@ describe('CustodianService', () => {
...
+      when(cacheManagerGet)
+        .calledWith(cacheId)
+        .mockReturnValueOnce({ requestId: custodianRequestId, status: ENUMS.FOLDER_STATUSES.CUSTODIAN_JOB_NOT_READABLE_YET });
+      when(httpServicePost)
+        .calledWith(...expectedHttpServiceReadJobArgs)
+        .mockReturnValueOnce(
+          httpResponseGenerator([
+            {
+              Started: CUSTODIAN.EMPTY_DATE,
+              Completed: CUSTODIAN.EMPTY_DATE,
```
oscar-richardson-softwire, line 211: Is the description for the test definitely right? It looks like the custodian job has empty date for both start and completion

src/modules/custodian/custodian.service.test.ts
```diff
@@ -93,8 +171,160 @@ describe('CustodianService', () => {
...
+    });
+
+    it('returns null if custodian job failed in previous run, and we will allow repeated call', async () => {
+      when(cacheManagerGet).calledWith(cacheId).mockReturnValueOnce({ requestId: custodianRequestId, status: ENUMS.FOLDER_STATUSES.CUSTODIAN_JOB_FAILED });
+
+      const response = await service.getApiResponseIfFolderInCustodian(folderParentId, folderName, res);
+
+      expect(res.status).not.toHaveBeenCalled();
+      expect(response).toBeNull();
+      expect(loggerWarn).toHaveBeenCalledWith('Found cache with failed Custodian job for folder creation. Lets try to create new job.');
```
oscar-richardson-softwire: `Lets try to create new job` -> `Let's try to create a new job`

src/modules/custodian/custodian.service.test.ts
```diff
@@ -93,8 +171,160 @@ describe('CustodianService', () => {
...
+      };
+      when(cacheManagerGet).calledWith(cacheId).mockReturnValueOnce({ requestId: custodianRequestId, status: ENUMS.FOLDER_STATUSES.SENT_TO_CUSTODIAN });
+      when(httpServicePost)
+        .calledWith(...expectedHttpServiceReadJobArgs)
+        .mockReturnValueOnce(throwError(() => axiosError));
+
+      const responsePromise = service.getApiResponseIfFolderInCustodian(folderParentId, folderName, res);
+
+      await expect(responsePromise).rejects.toBeInstanceOf(CustodianException);
+      await expect(responsePromise).rejects.toThrow(`Failed to get an Provisioning job by request id for cache id ${cacheId}.`);
```
oscar-richardson-softwire: `Failed to get an Provisioning job` -> `Failed to get a Provisioning job`

src/helpers/get-custodian-folder-creation-cache-key.helper.test.ts
avaitonis: Thank you, used variables to make calls more readable.

src/helpers/get-custodian-folder-creation-cache-key.helper.test.ts
avaitonis: Thanks, you description sounds better.

src/modules/custodian/custodian.service.test.ts
avaitonis: Made test more readable, but left same number of assessments. They test caching of one user journey, for me looks like they all are required.

src/modules/custodian/custodian.service.test.ts
avaitonis: Thanks, description didn't match test.

src/modules/custodian/custodian.service.test.ts
avaitonis: Thanks, changed

src/modules/custodian/custodian.service.test.ts
avaitonis: Thanks, changed

PR #581, opened 2024-05-16
feat(github-actions): enable various imperative github actions

abhi-markan:
## Introduction :pencil2:
Ensure various imperative pipelines are available for consistent pipeline actions across all the applications.

## Resolution :heavy_check_mark:
* Added `github.yml` for PR lint check.
* Added `prettier:all` check to `lint.yml`.
* Added `pr.yml` check for stale PR validation.
* Updated `publish.yml` as per latest `Google Release please` release notes.

## Miscellaneous :heavy_plus_sign:
* Dependencies updates
* Moved GHA hardcoded variables to Enterprise scope GH secrets / variable
* Refactored GitHooks


.github/workflows/deployment.yml
```diff
@@ -40,7 +40,7 @@ on:
 env:
   PRODUCT: apim
   ENVIRONMENT: ${{ github.ref_name }}
-  TIMEZONE: "Europe/London"
+  timezone: ${{ vars.TIMEZONE }}
```
avaitonis, line 43: Should we keep upper case env var name to be consistent?

avaitonis (approved):
LGTM, 
maybe upper case env var is not so consistent

.github/workflows/deployment.yml
abhi-markan: Thanks, have converted to lower case to keep it consistent with other projects will make them all uppercase in a separate PR.

